PR #5, opened 2019-04-05
fix(PF4 mapper): fixed pf4/react-core peer dependency.

Hyperkid123:
This should get rid of build warnings

### DIFF
--- a/packages/pf4-component-mapper/package.json
+++ b/packages/pf4-component-mapper/package.json
@@ -101,7 +101,7 @@
   },
   "peerDependencies": {
     "@data-driven-forms/react-form-renderer": "^1.9.3",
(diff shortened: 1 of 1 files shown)


rvsia:
:+1: 

@Hyperkid123 

There is also `@data-driven-forms/react-form-renderer": "^1.9.3` (same in other mappers)

Because of monorepo: should it not be `latest` or updated automatically (somehow)?

Hyperkid123:
> @Hyperkid123
> 
> There is also `@data-driven-forms/react-form-renderer": "^1.9.3` (same in other mappers)
> 
> Because of monorepo: should it not be `latest` or updated automatically (somehow)?

Don't know about that yet.

PR #21, opened 2019-05-15
feat(pf4wizard): update according to PF4

rvsia:
**Description**

- updates wizard according to PF4 component (however, it isn't using the component!)

**Props**

| Props  | Type | Default |  Decription |
| ------------- | ------------- | ------------- | ------------- |
| fields  | array  |   |   |
| isCompactNav  | bool  |   |  |
| setFullWidth  | bool  |   |   |
| setFullHeight  | bool  |   |   |
| isDynamic  | bool  |   |  |
| title  | node  |   |   |
| buttonsClassName  | string  |   |   |
| buttonLabels | object  |   |   |
| description  | node  |   |  |

**How to do substeps**

Field in Wizard fields should contain `substep` <`string`> which is title of the primary step. Steps with the same substep are combined together in the primary step.

**First step**

First step should have `stepKey: 1`

**Variants of Wizard**

**Simple wizard**

- steps navigation is visible, user can jump back and forward (after visiting the step)
- does not contain any conditional steps
- if you are need to change one step according to another (dynamic fields), do not use this (there is no way how to check which fields was changed), instead of that use **progressive wizard**
- if user change some field and make the form invalid, he cannot jump forward

simplewizard

**Progressive wizard**

- steps are visible as user visits them
- user can jump only back

progressivewizard

TODO
- [x] substep in progressive wizard
- [x] css styling (missing expanded option)
- [x] **tests**
- [x] doc for demo
- [x] demo css issue

### DIFF
--- a/packages/pf4-component-mapper/src/form-fields/wizard/wizard-styles.scss
+++ b/packages/pf4-component-mapper/src/form-fields/wizard/wizard-styles.scss
@@ -1,10 +1,7 @@
-.wizard-button-toolbar {
+.no-shadow {
+  box-shadow: none;
+}
+
+.pull-right {
   justify-content: flex-end;
-  margin-top: 1em;
-  button {
-    margin: 0 8px;
-    &:last-child {
-      margin-right: 0;
-    }
-  }
 }
--- a/packages/pf4-component-mapper/demo/demo-schemas/wizard-schema.js
+++ b/packages/pf4-component-mapper/demo/demo-schemas/wizard-schema.js
@@ -4,7 +4,11 @@ export const wizardSchema = {
   fields: [{
     component: componentTypes.WIZARD,
     name: 'wizzard',
+    //inModal: true,
     assignFieldProvider: true,
+    title: 'Title',
+    description: 'Description',
+    buttonsPosition: 'left',
     fields: [{
       title: 'Get started with adding source',
       name: 'step-1',
@@ -43,6 +47,7 @@ export const wizardSchema = {
       title: 'Configure AWS',
       name: 'step-2',
       stepKey: 'aws',
+      substepOf: 'Summary',
       nextStep: 'summary',
       fields: [{
         component: componentTypes.TEXT_FIELD,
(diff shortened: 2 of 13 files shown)


packages/pf4-component-mapper/src/form-fields/wizard/wizard-styles.scss
```diff
@@ -1,10 +1,7 @@
...
+.pull-right {
   justify-content: flex-end;
-  margin-top: 1em;
-  button {
-    margin: 0 8px;
-    &:last-child {
-      margin-right: 0;
-    }
-  }
 }
```
Hyperkid123: Can you add new line here? :wink:

rvsia:
@Hyperkid123 In the last rebase I just deleted some unused imports in wizardStep :skull_and_crossbones:

Hyperkid123:
@rvsia ok is it ready now?

Hyperkid123:
@rvsia also can you add some example of how the substeps are working? You kind of lost me there.

rvsia:
@Hyperkid123  Ok, give me a moment and I will add some examples to documentation. :+1:

rvsia:
I added this to demo:

<hr>

Example

Structure:

```jsx
                     //            index
Select Type          // step       1
Configuration        // step       2
  Security           // substep    2 (there is no step 'configuration')
  Credentials        // substep    3
Summary              // step       4
```

```jsx
Schema: [
  {
    stepKey: '1',
    title: 'Select Type',
    nextStep: 'sec'
  },
  {
    stepKey: 'security',
    title: 'Security',
    nextStep: 'credentials',
    substep: 'Configuration'
  },{
    stepKey: 'credentials',
    title: 'Credentials',
    nextStep: 'summ',
    substep: 'Configuration'
  },{
    stepKey: 'summary',
    title: 'Summary'
  },
]
```

Progressive Wizard works same way. It checks if previous step has the same \`substep\` value and if so, it grouped them together.
If the value is different, a new primary step is created with the step as a substep.

rvsia:
@Hyperkid123  changed substep prop to substepOf

However, I noticed that after rerender inputs lose focus :( I need to investigate, so WIP

Hyperkid123:
> However, I noticed that after rerender inputs lose focus :( I need to investigate, so WIP

what do you mean?

rvsia:
@Hyperkid123 when you type something in input fields, after first character it rerenders the wizard (probably because pristine is changed to false) and user needs to click on the input again. As I can see the first version in gif works correctly, I don’t know what went wrong.

PR #22, opened 2019-06-18
fix(mappers) expose pf3 and pf4 form mapper components.

Hyperkid123:
### Changes

Enabled directly importing form components. It is useful when trying to create new form component which is just a composite of existing ones. (https://github.com/ManageIQ/manageiq-ui-classic/pull/5614)


rvsia:
👍 looks great! But what about some documentation?  📖

Hyperkid123:
@rvsia yup and will do the same for PF4 too. Ill make this WIP and return to in once i have more time.

PR #26, opened 2019-07-09
fix(react renderer): Fixed length validators value assignment.

Hyperkid123:
There was a bug causing false value assignment to validator arguments. The OR statement caused the deprecated value to be always assigned.


rvsia:
What about to add some test which will check that the bug is no more?

PR #35, opened 2019-07-24
fix(pf4-select): Fixed clearing last value from multi select.

Hyperkid123:


packages/pf4-component-mapper/src/form-fields/select/select.js
```diff
@@ -145,11 +145,13 @@ export const Select = ({ selectVariant, ...props }) => {
         Input,
       }}
       { ...props }
-      onChange={ option => simpleValue
-        ? props.onChange(props.isMulti
-          ? option.map(item => item.value)
-          : option ? option.value : undefined)
-        : props.onChange(option) }
+      onChange={ (option) => {
+        let o =  !option && props.isMulti ? [] : option;
```
rvsia: :point_up:

Issue #47, opened 2019-07-26
Document component API


rvsia:
Document component API
Create a standard

PR #48, opened 2019-07-26
Adds Component API doc page

rvsia:
**Description**

Adds documentation for standard components API.

closes https://github.com/data-driven-forms/react-forms/issues/47

### DIFF
--- a/packages/react-renderer-demo/src/editor-demo/docs-components/component-api/component-api.md
+++ b/packages/react-renderer-demo/src/editor-demo/docs-components/component-api/component-api.md
@@ -1,0 +1,153 @@
+import ExampleLink from './exampleLink';
+
+### Components API
+
+Each mapper provided in `react-forms` provides a default API for standard components. Standard components are:
+<br />
+
+Wrapped in formGroup: <br/>
+&nbsp;&nbsp;text field <br/>
+&nbsp;&nbsp;text area <br/>
+&nbsp;&nbsp;select <br/>
+&nbsp;&nbsp;checkbox <br/>
+&nbsp;&nbsp;radio <br/>
+&nbsp;&nbsp;switch <br/>
+&nbsp;&nbsp;timepicker/datepicker <br/>
+<br />
+
+Others: <br/>
+&nbsp;&nbsp;subform <br/>
+&nbsp;&nbsp;tab/tab item <br/>
+&nbsp;&nbsp;wizard <br/>
+
+### FormGroup wrapped components
+
+Basic components are wrapped in formGroup. FormGroup provides a standard API to show labels, helper texts, errors, etc.
+
+All those components provides a shared group of props:
+
+|Prop|Type|Description|
+|----|:--:|----------:|
+|label|node/string|A label of the field|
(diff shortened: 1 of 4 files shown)


packages/react-renderer-demo/src/editor-demo/docs-components/component-api/component-api.md
```diff
@@ -0,0 +1,198 @@
...
+
+### FormGroup wrapped components
+
+Basic components are wrapped in formGroup. FormGroup provides a standard API to show labels, helper texts, errors, etc.
+
+All those components provides a shared group of props:
+
+| Prop        | Type           | Description  |
+| ------------- |:-------------:| -----:|
+| label     | node | A label of the field |
```
Hyperkid123: Is string also covered by node?

packages/react-renderer-demo/src/editor-demo/docs-components/component-api/component-api.md
```diff
@@ -0,0 +1,198 @@
...
+<br/>
+
+#### Checkbox
+
+| Prop        | Type           | Description  |
+| ------------- |:-------------:| -----:|
+| options     | array | Options in format { label: 'Label', value: value }, it will make it multiple a choice list. (optional)|
+
+<div style={{ float: 'right'}}>
+    <Button><Link component={NavLink} to={ `/component-example/checkbox` }>To example</Link></Button>
```
Hyperkid123: There should be link to both single and multiple chekcbox variant

packages/react-renderer-demo/src/editor-demo/docs-components/component-api/component-api.md
```diff
@@ -0,0 +1,198 @@
...
+| Prop        | Type           | Description  |
+| ------------- |:-------------:| -----:|
+| options     | array | Options in format { label: 'Label', value: value } |
+
+<div style={{ float: 'right'}}>
+    <Button><Link component={NavLink} to={ `/component-example/radio` }>To example</Link></Button>
+</div>
+<br/>
+
+#### Datepicker
```
Hyperkid123: Just making a note here, we don't have any API specified for PF4 just yet and MUI is probably not the same as for PF3.

rvsia: Yeah, I know. I just wanted to suggest that we could define some standard API for it and 
eventually re-implement it in other packages. What do you think? Or should I leave it for now just with the `placeholder` prop?

packages/react-renderer-demo/src/editor-demo/docs-components/component-api/component-api.md
rvsia: Should I add `string` to it?

packages/react-renderer-demo/src/editor-demo/docs-components/component-api/component-api.md
Hyperkid123: We will need it so we should think about the API. But there is no clue what kind of component will PF4 come up with.

PR #62, opened 2019-07-31
fix(pf4textfield): send number as a number

rvsia:


rvsia:
😿 closed because of https://github.com/data-driven-forms/react-forms/pull/67 😿

Issue #65, opened 2019-08-01
Data type conversion


Hyperkid123:
Convert field value data type if `dataType` attribute was specified.

Hyperkid123:
necessary for: https://github.com/ManageIQ/topological_inventory-ansible_tower/pull/17

PR #67, opened 2019-08-02
Data types feature

Hyperkid123:
closes: #65 

Adds data types casting to form fields (both default and custom)
- enhances FieldProvider `onChange` function with type casting if the `dataType` prop has been given
- supported data types
  - integer
  - float
  - number
  - boolean
  - string

### DIFF
--- a/packages/react-form-renderer/src/constants/index.js
+++ b/packages/react-form-renderer/src/constants/index.js
@@ -71,5 +71,5 @@ export const dataTypes = {
   FLOAT: 'float',
   NUMBER: 'number',
   BOOLEAN: 'boolean',
-  STRING: 'stirng',
+  STRING: 'string',
 };
--- a/packages/react-form-renderer/src/constants/index.js
+++ b/packages/react-form-renderer/src/constants/index.js
@@ -62,3 +62,14 @@ export const layoutComponents = {
   TITLE: 'Title',
   DESCRIPTION: 'Description',
 };
+
+/**
+ * @enum {String}
+ */
+export const dataTypes = {
+  INTEGER: 'integer',
+  FLOAT: 'float',
+  NUMBER: 'number',
+  BOOLEAN: 'boolean',
+  STRING: 'stirng',
+};
--- a/packages/react-form-renderer/src/form-renderer/enhanced-on-change.js
+++ b/packages/react-form-renderer/src/form-renderer/enhanced-on-change.js
@@ -1,0 +1,45 @@
+import { dataTypes } from '../constants';
+
+/**
+ * Pick a value from event object and returns it
+ * @param {Object|Any} event event value returned from form field
+ */
+const sanytyzeValue = event =>
+  (typeof event === 'object' && event.target && event.target.value)
+    ? event.target.value
+    : event;
+
+/**
+ * Casts string true/false to boolean
+ * @param {String} value value
+ */
+const castToBoolean = value => {
+  if (typeof value === 'boolean') {
+    return value;
+  }
+
+  return value === 'true';
+};
+
+/**
+ * Changes the value type
+ * @param {FieldDataTypes} dataType type for value conversion
+ * @param {Any} value value to be converted
+ */
+const convertType = (dataType, value) => ({
+  [dataTypes.INTEGER]: !isNaN(Number(value)) && parseInt(value),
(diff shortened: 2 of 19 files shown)


Hyperkid123:
cc @fryguy, @slemrmartin, @gmcculloug

packages/react-form-renderer/src/constants/index.js
```diff
@@ -62,3 +62,14 @@ export const layoutComponents = {
...
+
+/**
+ * @enum {String}
+ */
+export const dataTypes = {
+  INTEGER: 'integer',
+  FLOAT: 'float',
+  NUMBER: 'number',
+  BOOLEAN: 'boolean',
+  STRING: 'stirng',
```
rvsia: Typo here

Hyperkid123: where?

packages/react-form-renderer/src/form-renderer/enhanced-on-change.js
```diff
@@ -0,0 +1,45 @@
+import { dataTypes } from '../constants';
+
+/**
+ * Pick a value from event object and returns it
+ * @param {Object|Any} event event value returned from form field
+ */
+const sanytyzeValue = event =>
```
rvsia: sanitize

packages/react-form-renderer/src/validators/index.js
```diff
@@ -163,4 +163,7 @@ export const dataTypeValidator = type => ({
   number: options => pattern({
     pattern: /^\d*[.]{0,1}\d*$/, message: 'Values mut be number', ...options,
   }),
+  float: options => pattern({
+    pattern: /^\d*[.]{0,1}\d*$/, message: 'Values mut be number', ...options,
```
rvsia: mut > must be a number

packages/react-renderer-demo/src/editor-demo/docs-components/data-types.md
```diff
@@ -0,0 +1,13 @@
+import RawComponent from '../common/component/raw-component';
+
+## Introduction
+
+You can specify a type of a component by providing `dataType`, which will automatically validates the component value.
+Becuse almost everyting in html inputs is outputed as a string, adding the `dataType` props will also cast the value to given type.
```
rvsia: Because.. everything

packages/react-renderer-demo/src/editor-demo/docs-components/data-types.md
```diff
@@ -0,0 +1,13 @@
...
+
+### Avaiable dataTypes
```
rvsia: Available

rvsia:
Code looks solid! 👍 Just a few typos.. Need to test in UI. I like it, however, are we really make this feature as a default behavior? (To be honest, I haven't seen DataTypes used anywhere yet [except our demo schemas], but it can theoretically break some stuff if somebody uses it.)

Hyperkid123:
@rvsia it theoretically could break something but if yo are using data types and actually not expecting that data type in the output i don't thin it should be used at all. They can always remove it and just use validator.

packages/react-renderer-demo/src/editor-demo/docs-components/data-types.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Introduction
+
+You can specify a type of a component by providing `dataType`, which will automatically validates the component value.
+Because almost everything in html inputs is outputed as a string, adding the `dataType` props will also cast the value to given type.
+
+### Available dataTypes
+
+```jsx
+['integer', 'float', 'number', 'bool', 'string']
```
rvsia: boolean

martinpovolny:
Looks good!

packages/react-renderer-demo/src/editor-demo/docs-components/data-types.md
Hyperkid123: we will make this bool because Mozilla parser returns bool (i know we won't be using it). Let me fix it in the mapper.

rvsia: what about to allow both?

Hyperkid123: Nvm it is actually boolean

PR #79, opened 2019-08-07
Fixed md header anchor offset.

Hyperkid123:
Md anchor is no longer hidden below the top navbar.

fix found at: https://stackoverflow.com/questions/10732690/offsetting-an-html-anchor-to-adjust-for-fixed-header


rvsia:
:ice_cream: :+1:

PR #82, opened 2019-08-12
fix(renderer): fixed onChange for single checkbox.

Hyperkid123:
Form onChange was missing field type checking for checkboxes.


rvsia:
@Hyperkid123 please update snapshots :pray:

Issue #53, opened 2019-07-29
Async filtering in PF3/PF4 selects


rvsia:
Make async filtering in PF3/PF4 selects

cc @Hyperkid123

Hyperkid123:
In v2

PR #85, opened 2019-08-13
fix(pf4): Added async filtering to pf4 async select.

Hyperkid123:
Load options is called every time search input is changed. Options are cumulatively cached into component state.


rvsia:
Instead of 'loading data' it writes 'no options' :no_entry: 

Steps to reproduce in demo:
1. Wait for loading on mount
2. Write 'pepa'
3. Check the value

rvsia:
part of https://github.com/data-driven-forms/react-forms/issues/53 (just to link it together)

Issue #37, opened 2019-07-24
Document FormRenderer props


Hyperkid123:
There is no props description for FormRenderer

PR #88, opened 2019-08-15
Add renderer API doc

rvsia:
closes https://github.com/data-driven-forms/react-forms/issues/37


rvsia:
@Hyperkid123 FormSpy test was changed, I don't know why. Could you please check if it's ok?

Hyperkid123:
> @Hyperkid123 FormSpy test was changed, I don't know why. Could you please check if it's ok?

Probably the field array update. Its not critical it just sends the state update one more time..

Issue #98, opened 2019-09-03
Add name as a key to tabs component


Hyperkid123:
There are some problems with key value on pf3(possibly all) tabs component.

Hyperkid123:
cc @skateman

PR #100, opened 2019-09-03
Fix tabs

fhlavac:
closes: #98 

- Added name as a key to tabs component in PF3 and MUI.
- Modified priority of variables in adding name as a key to tabs component in PF4.
- Updated documentation of tabs component according to changes made.


Hyperkid123:
@fhlavac fix the failing snapshots please

PR #106, opened 2019-09-05
feat(pf4wizard): Allow custom buttons for steps

rvsia:
**Description**

Allow to use custom buttons components for each step (and the component is provided with all parts to be able to implement its own buttons)

**Schema**

```jsx
const Buttons = () => <div>Hello</div>;

[{
  title: 'foo-step',
  stepKey: '1',
  name: 'foo',
  buttons: Buttons,
  fields: [{
   name: 'foo-field',
  }],
  nextStep: '2',
}]
```

It will allow to do crazy things like this:

pf4wiz

TODO
- [x] Doc

### DIFF
--- a/packages/pf4-component-mapper/demo/demo-schemas/wizard-schema.js
+++ b/packages/pf4-component-mapper/demo/demo-schemas/wizard-schema.js
@@ -1,4 +1,32 @@
+import React, { useState } from 'react';
 import { componentTypes, validatorTypes } from '@data-driven-forms/react-form-renderer';
+import { Button } from '@patternfly/react-core';
+
+const ValidateButtons = ({ disableBack, handlePrev, buttonLabels: { back, cancel }, formOptions, renderNextButton }) => {
+  const [ state, setState ] = useState('init');
+
+  const setValidating = () => {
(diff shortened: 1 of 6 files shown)


packages/pf4-component-mapper/demo/demo-schemas/wizard-schema.js
```diff
@@ -1,4 +1,32 @@
+import React, { useState } from 'react';
 import { componentTypes, validatorTypes } from '@data-driven-forms/react-form-renderer';
+import { Button } from '@patternfly/react-core';
+
+const ValidateButtons = ({ disableBack, handlePrev, buttonLabels: { back, cancel }, formOptions, renderNextButton }) => {
+  const [ state, setState ] = useState('init');
```
Hyperkid123: no use state until react 17 is out :cry:

rvsia: Even not in demo? :sob:

Hyperkid123: oh its demo its fineee

PR #107, opened 2019-09-05
Focus search button when menu is opened

rvsia:
**Description**

Focus on search button when menu is opened


packages/react-renderer-demo/src/layout.js
```diff
@@ -105,9 +105,11 @@ const useStyles = makeStyles(theme => ({
 const Layout = ({ children, location: { pathname }}) => {
   const classes = useStyles();
   const [ open, setOpen ] = React.useState(false);
+  const searchRef = React.useRef(null);
 
   function handleDrawerOpen() {
     setOpen(true);
+    setTimeout(() => searchRef.current.focus(), 500);
```
Hyperkid123: Why the timeout? Was there some issue with immediate focus?

rvsia: Yes

PR #114, opened 2019-09-10
fix(Radio): Added default prop options to radio buttons.

Hyperkid123:
Added default prop options of empty array so the component does not throw an error. Has now same behavior like checkbox or select.

### DIFF
--- a/packages/pf3-component-mapper/src/form-fields/radio.js
+++ b/packages/pf3-component-mapper/src/form-fields/radio.js
@@ -1,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Radio } from 'patternfly-react';
+
+const RadioGroup = ({ FieldProvider, options, isDisabled, isReadOnly, input }) => options.map(option => (
+  <FieldProvider
+    key={ `${input.name}-${option.value}` }
+    name={ input.name }
+    value={ option.value }
+    type="radio"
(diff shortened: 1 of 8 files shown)


packages/pf3-component-mapper/src/form-fields/radio.js
```diff
@@ -0,0 +1,33 @@
...
+  })),
+  isDisabled: PropTypes.bool,
+  isReadOnly: PropTypes.bool,
+  input: PropTypes.shape({
+    name: PropTypes.string.isRequired,
+  }),
+};
+
+RadioGroup.defaultProps = {
+  options: [],
```
skateman: Is this the line I am interested in?

Hyperkid123: yes

skateman: then what's the sugar around it? it seems HUGE for me :confused:

Hyperkid123: You can assign default prop to anonymous function/component. 

Also if you notice all these three components in mappers are very similar. We plan to have "common" components that are shared between packages. This is just a step towards it.

skateman: Okay, it's a little complicated for me, so I think I'm not the right person to review this :disappointed:

PR #119, opened 2019-09-11
Fixed invalida textarea link.

Hyperkid123:
### Changes
- fixed invalida textarea link
- added fallback for non-existing components


packages/react-renderer-demo/src/pages/not-found-page.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+const NotFoundPage = () => {
```
rvsia: no need for `{}` block

Issue #52, opened 2019-07-29
Date / Time pickers API


rvsia:
I wanted to suggest that we could define some standard API for Date/Time pickers and
eventually re-implement it in other packages.

**Action after:** to update ComponentAPI doc page and packages' components.

PR #120, opened 2019-09-11
[WIP] Update date time picker

Hyperkid123:


rvsia:
If this closes #52, can you please also update the components API doc page?

Hyperkid123:
@rvsia still WIP. I need to add rest of the options (disabled from etc.)

rvsia:
@Hyperkid123 Oh sorry, didn't notice, I just clicked on requested review. My bad :crying_cat_face:

Hyperkid123:
there are some issue when switching between mappers and props with the date-picker (specifically isClearable and disabledDays). But it works in the specific demo for pf3 mapper. I think i will not add it to the docs example but i will add a link to the library we are using.

rvsia:
@Hyperkid123 I think that's okay, because PF4 Time/Datepicker is only placeholder (and MUI... :eyes: ). So we could come up with API now and then implement it in others mappers.

PR #121, opened 2019-09-12
Added prev/next links to doc pages.

### LINKED ISSUE #69: Add prev/next links to docs.
We should prev/next links to form renderer dos to make the guide more sense. Also we should think about the order of them and generally review them.

Hyperkid123:
closes #69

Added links


Hyperkid123:
@rvsia all issues should be fixed now.

PR #129, opened 2019-09-17
fix(ssr): added global object to renderer and pf4 webpack config.

Hyperkid123:
@rvsia can you please build it locally and test if it works with miq and probably some insights version.


rvsia:
|ManageIQ|SourcesUI|
|--------------|--------------|
|:ok:|:ok:|

PR #153, opened 2019-10-09
fix(all): migrated to rullop because of webpack issues.

Hyperkid123:
### Changes
- migrating to rollup for production builds
  - webpack was causing when minifying the output
  - we had several issues in the past with production builds so i think that the time has come to try and find different solution to these kinds of problems
  - webpack is still being used for local development

@rvsia please try this locally and link to the sources UI
@karelhala i know you had some issue in one of the insights app so please check it out
I did already tried it with PF3 mapper in miq-ui and the test issues seem to be resolved. Thanks guys.

cc @himdel @martinpovolny


karelhala:
Tested in FEC and everything works :+1:

PR #154, opened 2019-10-10
fix(pf3): async options reload

rvsia:
**Description**

Fixes re-loading when `options`, `loadOptions` are changed (and clears the input value if the new options do not include the value)

**After**
loadOptions

### DIFF
--- a/packages/pf3-component-mapper/src/form-fields/select/index.js
+++ b/packages/pf3-component-mapper/src/form-fields/select/index.js
@@ -2,7 +2,10 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import ReactSelect, { components } from 'react-select';
 import customStyles from './select-styles';
+import isEqual from 'lodash/isEqual';
 import './react-select.scss';
+
+const prepareFunction = (fn = '') => fn.toString().replace(/\s+/g, ' ');
 
 const ValueContainer = ({ children, ...props }) => {
   if (props.isMulti) {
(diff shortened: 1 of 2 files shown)


rvsia:
@Hyperkid123 

I am thinking about the value which stays during the reload:

What about to do...

when `options`) check if the value is in the new options and if not, set it to `undefined`

when `loadOptions`) 
   
- a) set it to `undefined` before the loading (we don't know the new options yet)

-    b) check after the loading

Hyperkid123:
@rvsia we should definitely check if the value is in the new options. Changin the Load options is easy we should just reset the value. Or if you want just check it after the first load that works too.

rvsia:
@Hyperkid123 Great! Will do it

packages/pf3-component-mapper/src/form-fields/select/index.js
```diff
@@ -41,10 +41,15 @@ class Select extends Component {
...
-    }));
+    .then((data) => {
+      if (!data.map(({ value }) => value).includes(this.props.input.value)) {
+        this.props.input.onChange(undefined);
+      }
+
+      return this.setState({
+        options: data,
+        isLoading: false,
+      });});
```
Hyperkid123: New line

rvsia:
changing

Hyperkid123:
Love it

Hyperkid123:
@skateman it should the problem you have with the schema update

skateman:
I'll test it soon

Issue #127, opened 2019-09-15
Difference between helperText and description


skateman:
The `description` and the `helperText` of an input field in PF3 is doing the exact same thing.

![Screenshot from 2019-09-15 15-21-39](https://user-images.githubusercontent.com/649130/64922261-891a9900-d7cd-11e9-82cc-a78bff8e63d2.png)

In the dialog editor there's no `description`, but that's actually fine. However, according to PF3 specs, the `helperText` should be rendered differently:
![Screenshot from 2019-09-15 15-27-36](https://user-images.githubusercontent.com/649130/64922267-99cb0f00-d7cd-11e9-9331-d11c0dc7dfcf.png)

Also if we're here, for consistency with English it should be `helpText` and not `helperText`.

Hyperkid123:
We will start working on soon

Hyperkid123:
cc @rvsia

Hyperkid123:
@rvsia do you know if there is something similar in PF4?

rvsia:
@Hyperkid123 PF4 design has also a `?` popover, but it isn't in the React implementation.https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms

Helper text is used to provide information to complete the form, the popover is used to provide background information.

Hyperkid123:
We have a bit of mismatch of naming here.

helperText in PF4 is the text under the input component but in PF3 its apparently the other way around. Since we are mapping PF4 API on PF3 components to have easy upgrade we are kind of screwed here.

skateman:
An alternative is to drop one of the two items and have just a single one, that will behave differently in PF3 and PF4.

skateman:
BTW, isn't the lack of [this](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms#provide-the-right-kind-of-help) a bug in pf-react?

Hyperkid123:
I think you shuold be able to compose it. There are necessary components i think but just not used together. @rvsia can correct me. He was playing with the popover in sources.

rvsia:
@Hyperkid123 Yeah, of course, you can compose it, but it's not support by FormGroup.

PR #156, opened 2019-10-10
fix(pf3): use fieldLevelHelp for helperText

rvsia:
closes https://github.com/data-driven-forms/react-forms/issues/127

`helperText` > icon popover
`description` > Help block

Before

image

After

image


rvsia:
@Hyperkid123 Will do

Hyperkid123:
@rvsia can you go through the manageiq and check the helper text usage?

We have to swap the attributes before we update to 1.19.x

rvsia:
@Hyperkid123 There is only one occurrence (in Workers form) and it should be definitely be swapped.

PR #157, opened 2019-10-11
fix(subform): allow to receive other props

rvsia:
**Description**

Allows subform to receive all props, it wasn't possible to add `className` before. 

I need to customize subforms in Sources.


Hyperkid123:
> I need to customize subforms in Sources.

@rvsia  Should't you customize it in sources then? I don't think that adding className is particularly disruptive that the customization is the whole point of this library.

PR #161, opened 2019-10-14
Use postcss for pf3

Hyperkid123:
### Issues
- wrong css build
  - using post css to be for css modules and injecting them
- named exports in rollup
  - i do not know yet how exactly combine named/default exports for node modules
  - we could only use named exports which would work but then there is an issue with jest, which is not using rollup internally. 
  - i also want to used build packages for tests to avoid issue with build packages in the future

we should deprecate all pf3 mappers between this PR and initial rollup PR.


packages/pf3-component-mapper/rollup.config.js
```diff
@@ -23,7 +23,11 @@ const babelOptions = {
...
 const commonjsOptions = {
   ignoreGlobal: true,
-  include: /node_modules/,
+  include: [
+    /node_modules/,
+    '../../node_modules/**',
+
+  ],
   namedExports: {
     '../react-form-renderer/dist/index.js': [ 'composeValidators' ],
```
rvsia: `    '../../node_modules/react-day-picker/DayPicker.js': [ 'DayPicker' ],`

PR #163, opened 2019-10-14
fix(pf3): give select menu higher z-index than input.

### LINKED ISSUE #160: Dropdown doesn't play well with date picker in PF3
Screenshot from 2019-10-11 18-21-35

Hyperkid123:
closes #160
For some reason normal form input with selector `.input-group .form-control` has z-index value of 2.

Since there is probably some reason for it and removing that might break everything, I just added higher value to select menu.


skateman:
I'm still experiencing the issue, but I might be doing something wrong. So let me play with the cache a little.

PR #165, opened 2019-10-14
fix(pf3): Date time picker placeholder.

### LINKED ISSUE #132: Same placeholder text for datepicker with date and date-time
The one with datetime should have datetime instead of date:
Screenshot from 2019-09-19 09-22-44

Hyperkid123:
closes #132

Added little different text to date-time picker default input placeholder.

### DIFF
--- a/packages/pf3-component-mapper/src/form-fields/date-time-picker/date-time-picker.js
+++ b/packages/pf3-component-mapper/src/form-fields/date-time-picker/date-time-picker.js
@@ -7,6 +7,12 @@ import { createDisabledDays } from './helpers';
 
 import 'react-day-picker/lib/style.css';
 
+const getPlaceholder = (variant, placeholder) => placeholder
+  ? placeholder
+  : variant === 'date'
+    ? 'Click in this input to select date'
+    : 'Click in this input to select date and time';
+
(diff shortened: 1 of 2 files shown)


packages/pf3-component-mapper/src/form-fields/date-time-picker/date-time-picker.js
```diff
@@ -7,6 +7,12 @@ import { createDisabledDays } from './helpers';
 
 import 'react-day-picker/lib/style.css';
 
+const getPlaceholder = (variant, placeholder) => placeholder
+  ? placeholder
+  : variant === 'date'
+    ? 'Click in this input to select date'
+    : 'Click in this input to select date and time';
```
skateman:
```suggestion
    ? 'Click in this input to select a date'
    : 'Click in this input to select a date and a time';
```

I'm not sure about the second `a` in the `a date and a time` part, maybe we need someone who actually speaks English :laughing:

d-m-u: They're both okay from a grammatical standpoint, both the original and the suggestion.

Issue #171, opened 2019-10-18
Consistency in top-level attributes of the schema


skateman:
I noticed that if you set a `description` on the top level of the schema, it gets rendered consistently with descriptions on lower levels. However, the `label` attribute is not getting rendered. 

So for example, the following schema:
```js
{
  label: 'Dialog editor demo',
  description: 'This is a demo of the new dialog editor that is based on Data Driven Forms',
  fields: [...]
}
```

Gets rendered like this:
![Screenshot from 2019-10-18 13-07-00](https://user-images.githubusercontent.com/649130/67089667-81715a00-f1a8-11e9-8ecb-71a48e925af6.png)

To be honest, I'm positively surprised that the `description` is rendered and I don't have to implement it, but I think the `label` should be rendered as well. However, I would totally understand if you don't want to render the `label`, but in this case you would have to do similarly with the `description`.

rvsia:
Yeah, there is `title` prop, which works like that, but it makes sense to have `label` too to be consistent. Check the PR I made, I also updated docs to get this info clear. Thanks for the issue!

PR #172, opened 2019-10-18
fix(renderer): add label prop to form

rvsia:
closes https://github.com/data-driven-forms/react-forms/issues/171

**Description**

- adds `label` prop to form, same as `title`
- updates docs


skateman:
@rvsia it's a minor problem and I'm not a designer, but IMO the top-level label should be bigger than the label on the sub-form. With the same analogy as the individual labels for inputs are smaller than labels for sub-forms.

```
top-level > sub-form > input
h1 > h3 > label
```

rvsia:
@skateman I agree, this was added only as a placeholder.

https://github.com/data-driven-forms/react-forms/blob/5758020811f9b8cbaae89bfd2920553c2b1dc08b/packages/pf3-component-mapper/src/form-fields/layout-components.js#L37 Here, you can change it.

PR #180, opened 2019-10-22
Allow array value select

### LINKED ISSUE #179: Multiselect accepts non-array initialValue, single select doesn't accept array
There's a slight inconsistency in how `initialValue` is handled between single and multiselect. Would it be possible for the single select to accept an array of one element?

Hyperkid123:
closes #179 

### Changes
- added new prop `pluckSingleValue` for pf3 select
- you can now use array as a value for single select. It will pick the value of first item in array
- `[2, 4, 5]` will result into `2` as a value in form state
- also its now possible to use an array as a select value
```jsx
[{
  label: 'Foo'
  value: [2, 4, 5]
}]
```


skateman:
@Hyperkid123 does this mean that I have to use the `pluckSingleValue`? What happens if I use it on a multiselect?

Hyperkid123:
@skateman  `pluckSingleValue` is by default `true` and is ignored on multi select.

packages/pf3-component-mapper/src/form-fields/select/index.js
```diff
@@ -164,6 +170,7 @@ Select.defaultProps = {
   },
   loadingMessage: 'Loading...',
   simpleValue: true,
+  pluckSingleValue: true,
```
rvsia: This could bring some incompatibility issue, if someone uses an actual array as a value for the select.

skateman:
Okay, then I just have to try it :wink:

packages/pf3-component-mapper/src/form-fields/select/index.js
Hyperkid123: Yep i though of that. Check the rest of the code and tests :wink:

packages/react-renderer-demo/src/docs-components/pf3-select.md
```diff
@@ -1,3 +1,11 @@
+### Select
+
+There are some additional props for PF3 variant to match PF3 requirements.
+
+|prop name|type|description|
+|---------|----|-----------|
+|pluckSingleValue|`bool`|If a value is an array, component will pick its first item and set it as a new value. This will override the value in state from `[2, 4, 5]` to `2` for example.|
```
rvsia: It should be said that it is a `true` by default

packages/pf3-component-mapper/src/form-fields/select/index.js
rvsia: I thought that if someone has options like this right now:

```jsx
[
  { value:  ['apple', 'pineapple'], label: 'sweet'},
  { value: ['chocolate','kitkat'], label: 'best'},
] 
```

Then he should have to set this prop to `false` and that will break the backwards compatibility. Am I right?

Hyperkid123: The example you posted would not work at this point because we were using `===` to select the value from options instead if `isEqual`. So nobody could use it successfully

Issue #176, opened 2019-10-18
Higher order components with clearOnUnmount ignore initialValue


skateman:
If an item in the schema with `clearOnUnmount` is implemented using a higher order component, the rendered field ignores the `initialValue` of the given element. First I thought it's a side-effect of my `Options` component, but I was able to narrow it down to the fact that this is caused by having it as a HOC.

PR #181, opened 2019-10-22
fix(pf3): fixed initial value for pf3 date-picker.

### LINKED ISSUE #177: Datepicker keeps displaying placeholder if initialValue is set
The datepicker doesn't display the `initialValue`, but the placeholder instead.

Hyperkid123:
closes #177 

### Changes
- fixed initial value not being propagated to pf3 date picker.


skateman:
It works, but I keep bumping into https://github.com/data-driven-forms/react-forms/issues/176

PR #188, opened 2019-10-31
Firebase fix

rvsia:


rvsia:
@Hyperkid123 Wizard demo is missing its modified formFieldsMappers.

Otherwise it looks that everything works as it should.

Missing pages/features I noticed:
- field array provider page
- PF4 wizard docs
- PF3/PF4 selects docs
- form renderer api

PR #190, opened 2019-10-31
fix(pf4wizard): fix predicting steps for func nextStep

rvsia:
**Description**

Fixes predicting steps and disabling the nav, when the nextstep is a function.

- plus refactors tests and adds a few new use cases

Todo:

- [x] tests

### DIFF
--- a/packages/pf4-component-mapper/src/form-fields/wizard/wizard.js
+++ b/packages/pf4-component-mapper/src/form-fields/wizard/wizard.js
@@ -8,6 +8,8 @@ import get from 'lodash/get';
 import set from 'lodash/set';
 import flattenDeep from 'lodash/flattenDeep';
 
+const DYNAMIC_WIZARD_TYPES = [ 'function', 'object' ];
+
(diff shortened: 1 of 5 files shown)


packages/pf4-component-mapper/src/form-fields/wizard/wizard.js
```diff
@@ -19,7 +21,9 @@ class Wizard extends React.Component {
     super(props);
 
     // find if wizard contains any dynamic steps (nextStep is mapper object)
-    const isDynamic = this.props.isDynamic ? true : this.props.fields.find(({ nextStep }) => typeof nextStep === 'object') ? true : false;
+    const isDynamic = this.props.isDynamic ? true : this.props.fields.find(({ nextStep }) => DYNAMIC_WIZARD_TYPES.includes(typeof nextStep))
+      ? true
```
Hyperkid123: :1st_place_medal: This reminds us some code from ManageIQ. I dont think we need ternary operator here. Maybe use `!!find()...`

rvsia: :face_with_thermometer:

PR #191, opened 2019-11-01
fix(pf3): stop date picker crash when value is string.

### LINKED ISSUE #189: Opening a datepicker with a preset initialValue crashes
The schema below throws an an `e.getTime is not a function` error:
```javascript
{
  component: componentTypes.DATE_PICKER,
  name: "date-picker-1",
  label: "Datepicker 1",
  initialValue: "2019-11-01T12:31:00.000Z",
  disabledDays: [
    {
      before: "today"
    }
  ],
  variant: "date"
}
```

Hyperkid123:
closes #189 

Day picker can now accept value as a string.

### DIFF
--- a/packages/pf3-component-mapper/demo/demo-schemas/sandbox.js
+++ b/packages/pf3-component-mapper/demo/demo-schemas/sandbox.js
@@ -431,6 +431,7 @@ const output = {
                   title: 'Datepicker with past days',
                   component: components.DATE_PICKER,
                   variant: 'date-time',
+                  initialValue: '2019-11-04T12:31:00.000Z',
                 },
               ],
               component: components.SUB_FORM,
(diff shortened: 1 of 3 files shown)


Hyperkid123:
ups forgot add the WIP thing Working on it right now.

skateman:
Now it's throwing me `Invalid Date` if the `initialValue` is not set:
```json
{
  "component": "date-picker",
  "name": "date-picker-1",
  "label": "Datepicker 1",
  "visible": true,
  "disabledDays": [
    {
      "before": "today"
    }
  ],
  "variant": "date"
}
```

Screenshot from 2019-11-01 10-15-29

And by clicking on the picker icon, it crashes with `Cannot read property 'length' of undefined`

Hyperkid123:
@skateman ah there is one mor thing i forgot

rvsia:
@Hyperkid123 what about to add a test that proves that `picker icon click crash` is gone, just to be sure it won't repeat in the future? Otherwise looks good!

Hyperkid123:
@rvsia the crash has nothing to do with the clicl but with the fact the it was sending string to the day picker component instead of Date object.

PR #198, opened 2019-11-04
fix(pf3): fixed date picker initialization.

Hyperkid123:
fixes: https://github.com/skateman/ddf-editor/issues/2

If field is pristine internal state is overridden with props value

### DIFF
--- a/packages/pf3-component-mapper/src/form-fields/date-time-picker/date-time-picker.js
+++ b/packages/pf3-component-mapper/src/form-fields/date-time-picker/date-time-picker.js
@@ -31,7 +31,7 @@ export class DateTimePicker extends React.Component {
     super(props);
     this.state = {
       positionLeft: 0,
-      selectedDay: props.value ? typeof props.value === 'string' ? new Date(props.value) : props.value : undefined,
+      selectedDay: this.createDate(),
       selectingMonth: false,
       selectingYear: false,
       isOpen: false,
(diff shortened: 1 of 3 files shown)


packages/pf3-component-mapper/src/tests/form-fields/date-time-picker/date-time-picker.test.js
```diff
@@ -18,5 +18,22 @@ describe('<DateTimePicker />', () => {
...
   });
+
+  it('should initialize date picker value if pristine is set to true', () => {
+    const initialDate = '2019-11-01T12:31:00.000Z';
+    const wrapper = mount(<DateTimePicker value={ initialDate }/>);
+    expect(wrapper.state().selectedDay).toEqual(new Date(initialDate));
+    wrapper.setProps({ pristine: false });
+    wrapper.update();
+
+    wrapper.setProps({ value: '2029-11-01T12:31:00.000Z' });
```
rvsia: Can you set both props at the same time?  Also, I would extract `'2029-11-01T12:31:00.000Z'` to separate const and use it here and in the equal.

packages/pf3-component-mapper/src/form-fields/date-time-picker/date-time-picker.js
```diff
@@ -45,6 +45,18 @@ export class DateTimePicker extends React.Component {
     });
   }
 
+  componentDidUpdate({ pristine }) {
+    if (this.props.pristine === true && pristine === false) {
+      const date = this.props.value ? typeof this.props.value === 'string' ? new Date(this.props.value) : this.props.value : undefined;
```
rvsia: this is used at two places now, I would extract this function to make additional modifications easier.

skateman:
I see no change in the editor with the linked branch :confused:

Hyperkid123:
@skateman it does work. Is it possible that you are not sending pristine value?
initialize

skateman:
@Hyperkid123 this is how I'm sending it https://github.com/skateman/ddf-editor/blob/master/src/Draggable/Input.jsx#L13 and it works with every other input field...

Hyperkid123:
@skateman datepicker requires `pristine` prop to override its internal state. Pristine comes from FieldProvider and every other field is using its. It comes from the `meta` object from FieldProvider.

It has nothing to do with `initialize` call.

Hyperkid123:
@skateman i will try and get rid of the internal state. I remember that there were some issues with time. But don't know exactly why it was there. That should resolve the requirement of `pristine` prop and solve the issue as well. Might not be able to do it though.

Hyperkid123:
@skateman ok try it now. Internal state for value is gone. It should be always synchronized with value from form state as are the other fields.

PR #199, opened 2019-11-04
Multi field validation

### LINKED ISSUE #149: Multi-field validation
I noticed in the provider forms, that some of them have a little bit more strict schema than in our current implementation. More specifically, filling the capacity and utilization data is strictly mandatory in them. However, in our current implementation, we make it mandatory only if at least one of the fields in that section is filled out.

Could we do something similar in DDF? Let's say if you fill out a single field in a SUB_FORM, some other fields in that SUB_FORM become mandatory.

cc @Fryguy

Hyperkid123:
closes #149

### DIFF
--- a/packages/react-renderer-demo/src/app/pages/renderer/condition.md
+++ b/packages/react-renderer-demo/src/app/pages/renderer/condition.md
@@ -34,7 +34,7 @@ You can show a field only if it meets a condition:
 
 `when` - is name of field where the value is stored, **always required!**. It can be either string `'field-name'` or array of strings `['field-1', 'field-2']`.
 
-#### OR condition. At least one condition must be met
+### OR condition. At least one condition must be met
 
 If either of of fields with name `a` and `b` will have value `x` condition is met.
 
@@ -51,7 +51,7 @@ If either of of fields with name `a` and `b` will have value `x` condition is me
 }
 ```
 
-#### AND condition. All conditions must be met
+### AND condition. All conditions must be met
 
 Field `controlled-field-1` must have value `Bar` and field `controlled-field-2` must include `FooBar` somewhere in its value to display field `BarFoo`.
 
--- a/packages/react-renderer-demo/src/app/pages/renderer/condition.md
+++ b/packages/react-renderer-demo/src/app/pages/renderer/condition.md
@@ -32,7 +32,54 @@ You can show a field only if it meets a condition:
 }
 ```
 
-`when` - is name of field where the value is stored, **always required!**
+`when` - is name of field where the value is stored, **always required!**. It can be either string `'field-name'` or array of strings `['field-1', 'field-2']`.
+
+#### OR condition. At least one condition must be met
+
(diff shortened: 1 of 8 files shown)


skateman:
Would be nice to have some kind of documentation about how to use this.

packages/react-renderer-demo/src/docs-components/condition.md
```diff
@@ -25,7 +25,33 @@ You can show a field only if it meets a condition:
...
+    } {
+      name: 'BarFoo',
+      label: 'Foo is Bar!',
+      component: 'text-field',
+      condition: [{
+        when: 'controlled-field-1',
+        is: 'Bar',
+      }, {
+        when: 'controlled-field-2',
+        patter: /FooBar/
```
skateman:
```suggestion
        pattern: /FooBar/
```

packages/react-renderer-demo/src/docs-components/condition.md
```diff
@@ -25,7 +25,33 @@ You can show a field only if it meets a condition:
 }
 ```
 
-`when` - is name of field where the value is stored, **always required!**
+`when` - is name of field where the value is stored, **always required!**. It can be either string `'field-name'` or array of strings `['field-1', 'field-2']`. Conditions are using **OR** operator. If you want **AND** you can pass multiple conditions.
```
skateman: >  Conditions are using **OR** operator. If you want **AND** you can pass multiple conditions.

This is really hard to understand :confused:

Hyperkid123: I added explicit examples. I would be grateful for any suggestions.

skateman:
@Fryguy @agrare this seems as a viable solution for optional but if one filled out then required fields in provider forms.

rvsia:
@Hyperkid123 Docs update :book: :fire:

packages/react-renderer-demo/src/app/pages/renderer/condition.md
```diff
@@ -32,7 +32,54 @@ You can show a field only if it meets a condition:
 }
 ```
 
-`when` - is name of field where the value is stored, **always required!**
+`when` - is name of field where the value is stored, **always required!**. It can be either string `'field-name'` or array of strings `['field-1', 'field-2']`.
+
+#### OR condition. At least one condition must be met
```
rvsia: Should be only three levels ###

packages/react-renderer-demo/src/app/pages/renderer/condition.md
```diff
@@ -32,7 +32,54 @@ You can show a field only if it meets a condition:
...
+    component: 'text-field',
+    condition: {
+      when: ['a', 'b'],
+      is: 'x'
+    }
+  }]
+}
+```
+
+#### AND condition. All conditions must be met
```
rvsia: Should be only three levels ###

rvsia:
@Hyperkid123 There is a minor documentation inconsistency (since this PR we have rewritten all header's levels)

PR #209, opened 2019-11-06
Make nextdocs consistent with master

rvsia:


rvsia:
@Hyperkid123 Next docs should be updated to master!

PR #218, opened 2019-11-11
Adjusted doc layout to comply with algolia doc-search.

Hyperkid123:
Added algolia requirements https://community.algolia.com/docsearch/required-configuration.html


rvsia:
Codecov is somewhat broken here...

PR #233, opened 2019-11-18
Ignore empty values length validator

### LINKED ISSUE #231: Provide a way how to ignore empty values in Length validators
**Description**

In Sources-UI I am using the same schema for adding/editing a source and ARN has to have minimal character count. However, because it is a 'password' value, I don't get the value from the API I can't repopulate the schema, so I cannot edit a source, because it is empty.

And even all other validators ignore empty values. 

So I think we should do either:

A) Ignore empty values 
or
B) Add an additional option to ignore empty values

I would love to do the first one (consistency and if user needs to have some value, he can always use the required validator), but it could bring breaking changes, so I think we should go the second way.

What do you think, @Hyperkid123?

rvsia:
Closes #231

### DIFF
--- a/packages/react-form-renderer/src/tests/validators/validators.test.js
+++ b/packages/react-form-renderer/src/tests/validators/validators.test.js
@@ -28,6 +28,14 @@ describe('New validators', () => {
   });
 
   describe('Length validator', () => {
+    it('should pass empty value', () => {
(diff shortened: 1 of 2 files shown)


Hyperkid123:
@rvsia commit message? I think this should be released.

PR #237, opened 2019-11-19
Fixed demo links to enable crawlers.

Hyperkid123:
- Moved the docsearch class name to correct element
- added links to anchors to enable site crawling


packages/react-renderer-demo/src/app/src/components/mdx/mdx-components.js
```diff
@@ -21,13 +22,14 @@ const CodeEditor = dynamic(import('../code-editor'), {
 
 const useHeadingStyles = makeStyles(() => ({
   anchorOffset: {
+    textDecoration: 'none',
+    color: 'inherit',
+    paddingTop: 92, // compensate for fixed header size and spacing
```
rvsia: Could we make a constant for this?

PR #248, opened 2019-11-26
fix(icons): use @patternfly/react-icons

redallen:


Hyperkid123:
@redallen thank you. Bundle sizes are same as before. I would like to continue discussion about this in referenced issue.

PR #260, opened 2019-12-10
fix(pf4 select): Added prop that enables select menu portaling

Hyperkid123:
Use `menuIsPortal` prop to enable it. portal is appended to body element.

### Before
Screenshot from 2019-12-10 13-12-29

### After
screenshot-localhost_8080-2019 12 10-13_00_50


packages/react-renderer-demo/src/app/src/doc-components/pf4-select.md
```diff
@@ -1,3 +1,11 @@
+# Menu portaling
+
+In order to show menu properly in dialogs like Modals, you can use `menuIsPortal`. This will append portal to body to render menu.
```
rvsia: `modals` with small m?

`This will append portal to body to render menu.` Is this sentence right?

rvsia:
@Hyperkid123 Tests failed

PR #262, opened 2019-12-10
Split component example into separate files.

Hyperkid123:
Use full routing instead of params in component examples to enable it for docs indexing.

### DIFF
--- a/packages/react-renderer-demo/src/app/src/components/navigation/examples-definitions.js
+++ b/packages/react-renderer-demo/src/app/src/components/navigation/examples-definitions.js
@@ -7,7 +7,7 @@ import SelectText from './examples-texts/select';
 
 export const baseExamples = [{
   component: componentTypes.TEXT_FIELD,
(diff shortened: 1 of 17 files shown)


packages/react-renderer-demo/src/app/src/components/navigation/examples-definitions.js
```diff
@@ -7,7 +7,7 @@ import SelectText from './examples-texts/select';
 
 export const baseExamples = [{
   component: componentTypes.TEXT_FIELD,
-  link: `?component=${componentTypes.TEXT_FIELD}`,
+  link: componentTypes.TEXT_FIELD,
```
rvsia: links and components are now the same, we should use only one

Hyperkid123: Yeah. Links and component are a bit different. I would like them to be separate. It was like that before the srr switch and it was really confusing.

PR #267, opened 2019-12-12
Use hash when searching

rvsia:
This appends hash to algolia suggestions.

### DIFF
--- a/packages/react-renderer-demo/src/app/src/components/docsearch.js
+++ b/packages/react-renderer-demo/src/app/src/components/docsearch.js
@@ -54,7 +54,7 @@ const DocSearch = () => {
       handleSelected: (input, event, suggestion) => {
         event.button = 0;
         const resultUrl = new URL(suggestion.url);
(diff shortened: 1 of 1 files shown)


packages/react-renderer-demo/src/app/src/components/docsearch.js
```diff
@@ -54,7 +54,7 @@ const DocSearch = () => {
       handleSelected: (input, event, suggestion) => {
         event.button = 0;
         const resultUrl = new URL(suggestion.url);
-        push(resultUrl.pathname);
+        push(`${resultUrl.pathname}${resultUrl.hash}`);
```
Hyperkid123: what if the hash is undefined?

rvsia: Can it be?

Hyperkid123: Well thet depends on the algolia crawler. We are not in control here

PR #272, opened 2019-12-18
Use select componsition to style pf3 select

Hyperkid123:
### Changes
- use traditional css styling instead of css in js
- use composite components
- the searchable variant is now using PF dropdown as a base of the select to avoid positioning the input by css
- main logic was moved to common package
- async has to stay package-specific to due to the dropdown base (sorry can't do anything about that) and constant re-initializing if the component in searchable variant
- fixed some PF styling inconsistency 

@rvsia @skateman this was a lot of refactoring and lof of the internal logic has changed. I would appreciate it if you could test this very thoroughly since there is a high chance that I've missed something. Even though the tests might pass I would like to test it manually.


rvsia:
@Hyperkid123 Looks great, however, the code coverage decreased too much. Could you add some test to increase it please?

Hyperkid123:
i will once i am able

PR #273, opened 2019-12-19
Cleared value prop

Hyperkid123:
cc @skateman @himdel 

If you come up with a better name for this please help.


skateman:
What does this do exactly?

Hyperkid123:
@skateman so when you have a form for editing and you delete some value which had an initial value, after submitting it will be `null` (or something else you want) in the state. If you don't set it will be `undefined`

We talked about it some time ago and told you that I don't think it should be here. Well you were right :wink:

rvsia:
What about `emptyValue`, `clearedValue`?

packages/react-form-renderer/src/form-renderer/enhanced-on-change.js
```diff
@@ -40,19 +41,45 @@ const convertType = (dataType, value) => ({
   [dataTypes.BOOLEAN]: castToBoolean(value),
 })[dataType] || value;
 
+/**
+ * Checks the value and returns undefined if its empty. Converst epmtry strings, arrays and objects.
```
himdel: epmtry

packages/react-form-renderer/src/form-renderer/enhanced-on-change.js
```diff
@@ -40,19 +41,45 @@ const convertType = (dataType, value) => ({
...
+ * If value is empty its overriden to undefined for further processing.
+ * @param {Any} value Any JS variable to be check if is empty
+ */
+const checkEmpty = value => {
+  if (typeof value === 'number') {
+    return value;
```
himdel: 0 is a valid number, yet, `if (checkEmpty(0))` is falsy. Probably not what you want?

himdel: Maybe what I'm really saying is that `checkEmpty` should return true/false, and the `typeof checkEmpty(result) === "undefined"` check should just be `checkEmpty(result)`.

Hyperkid123: Is it? There is actually a test with `0` which does return the correct value `0`, not the deleted value.

https://github.com/data-driven-forms/react-forms/pull/273/files#diff-e027b205eabea394b0be42c72f4e7681R49

Maybe I am missing something.

himdel: Yeah, I did not expect the test to be like that, given the name of the function.
It does work as is, it's just surprising :).

himdel:
To me, `emptyValue` makes the most sense,
especially if we also start using it to initialize fields without `initialValue`.

packages/react-form-renderer/src/form-renderer/enhanced-on-change.js
Hyperkid123: Yeah, you might be right. All that 0, -1, etc being false makes me uncomfortable using boolean in this case. Returning boolean directly makes it cleaner.

Hyperkid123:
> (not sure if that^ would break something?)

It won't thanks.

rvsia:
I would rather skip the using of this value to initialize values.

My current use-case is that I have multiple API calls on a submit and I want to call only endpoints which are really needed - which is decided by naming of form values.

Current state:
```
authentication.password: undefined,
credentials.name: null,
```

all `authentication.*` values are going to a authentication API
all `credentials.*` are going to a credentials API

So in this case, the authentication API is not called, only the credentials API is and the name is deleted.

If we used this props to initialize values, it would end in sending empty password to the authentication as well.

I suggest if we want to be able to set up a global initial value, let's create another form renderer prop and call it `initialValue`.

packages/react-form-renderer/src/form-renderer/field-provider.js
```diff
@@ -28,7 +28,9 @@ class FieldProvider extends Component{
           { ...fieldsProps }
           input={{
             ...input,
-            onChange: (...args) => enhancedOnChange(dataType, onChange, ...args),
+            onChange: (...args) => {
```
rvsia: I don’t think this will cause any issue, but using the block you removed the return statement.

I would not use this block, at least for code consistency.

himdel:
So, @rvsia, I don't care about having the initialValue initailized like this, the suggestion was only for the name to make sense. So, fine with not having it, but in that case, I guess `deletedValue` is right.

But, what you describe sounds wrong... if you need to actually distinguish untouched fields from changed fields, that's a job for metadata, not for the actual field values. 

Maybe what we really need is an array of pristine/dirty fields?

Don't make assumptions about the meaning of null **or** undefined, these can be completely valid values for some kinds of inputs.

packages/react-form-renderer/src/form-renderer/enhanced-on-change.js
```diff
@@ -40,19 +41,45 @@ const convertType = (dataType, value) => ({
...
-  return onChange(
-    Array.isArray(sanitizedValue)
-      ? sanitizedValue.map(item => convertType(dataType, sanitizeValue(item)))
-      : convertType(dataType, sanitizedValue),
-    ...args);};
+  const result = Array.isArray(sanitizedValue)
+    ? sanitizedValue.map(item => convertType(dataType, sanitizeValue(item)))
+    : convertType(dataType, sanitizedValue);
+  if (typeof checkEmpty(result) === 'undefined' && typeof initial !== 'undefined') {
+    return onChange(deletedValue, ...args);
```
himdel: Actually this may be problematic...

even if we don't use `deletedValue`, this will now always convert "" to `undefined`, right?

I think `""` and `undefined` are **very** different, unless the fields has a very specific meaning.

himdel:
> I would rather skip the using of this value to initialize values

Oh, maybe I misunderstood though.
Because, even if we use `emptyValue` to initialize **missing** initialValues, you can still just initialize it to `undefined` first.

Remember the difference between `key in object` and `object[key] === undefined` ;).

himdel:
Sorry for the spam but one more random thought: is `emptyValue` truly something that can be form-global?

Wouldn't it be better as a field property instead?

himdel:
> authentication.password: undefined,
> credentials.name: null,

@rvsia What if the user starts entering a password, changes their mind, and deletes it?
You'll get `null` in both, but the situation should be the same. That's why I think that approach sounds wrong.

himdel:
(If the whole idea is just for deletable password fields, a component with an explicit button which just sets value to null may be a better solution.)

The more I think of it, DDF should make as little assumptions about the meaning of field values as possible, and there's no common "deleted value" that will work for any single field type, so unless I'm missing something, this should be field logic, not form logic.

rvsia:
@himdel 

> Maybe what we really need is an array of pristine/dirty fields?

We have these. However, it only computes values, which are currently present in DOM (behavior of external Final Form library we are using), so it's not very useful for wizard, where I need it. That was I was told @Hyperkid123 that we should introduce this prop, because otherwise I have to do complex post-processing of the wizard values to get know, which data were removed, updated, added or just touched.
 
> can be form-global? Wouldn't it be better as a field property instead?

We can have both options, but the second one means to rewrite all schemas. The first means to rewrite only the renderer component.

I would also like to introduce an option to put there a function, where user can do anything he likes to do. (i.e.: Is it select? Put `null`, is it text? put `''` etc.)

> What if the user starts entering a password, changes their mind, and deletes it?

Initial state
```jsx
credentials.name: 'some name',
```

Final state
```jsx
authentication.password: undefined,
credentials.name: null,
```

In this case, I still know what the user changed, so it's working perfectly for me.

packages/react-form-renderer/src/form-renderer/enhanced-on-change.js
Hyperkid123: `"" => undefined` is being done anyway by the form API. You can specify `parse` prop in field to change it. So this was happening even without the deleted value feature.

Hyperkid123:
@himdel @rvsia I am officially very confused about all these comments :smile:. Can we can some input-output of what would you expect from this?

Hyperkid123:
> Sorry for the spam but one more random thought: is `deletedValue` (or `emptyValue`, whatever) truly something that can be form-global?
> 
> Wouldn't it be better as a field property instead?
> 
> (Either that, or a well-defined interface for a custom component to say what is actually "deleted".)

We can do it both ways as with `clearOnUnmount`. You can specify if globally and the in field. The field specification has always higher priority.

himdel:
:+1: thanks both of you, as far as I can tell, I'm not seeing any issues then :)

packages/react-renderer-demo/src/app/pages/renderer/deleted-value.md
```diff
@@ -0,0 +1,20 @@
...
+import RawComponent from '@docs/raw-component';
+
+import ListOfContents from '../../src/helpers/list-of-contents';
+
+<Grid container item>
+<Grid item xs={12} md={10}>
+
+# Deleted value
+
+If you required a specific value for a field which had **initialValue** to set it to some specific value. For instance if you have a form which edits some entity and you need to delete some attribute, APIs usually require the value to be set to `null` to register the change.
```
skateman: Maybe it's just me, but the first sentence doesn't make any sense :confused: I can probably guess that the field will send some value if make it empty, which in our cases needs to be `null`. If that's the case, I'd name it `emptyValue`, `clearValue` or `clearedValue` and the description could be something like this:

> The value to send upon a submit if the field is empty. This wildcard value can be used to distinguish between an untouched field and a cleared one. For example if you have a form that edits an entity and you would like to clear an attribute. Some APIs require the value to be set to `null` to register the change.

What do you guys think?

PR #277, opened 2020-01-09
Use tranform pluging to transform PF4 icons

rvsia:
Not sure I am doing it right. :dango: 

`react-icons/dist/umd || js || esm/icons/` we don't need use case here for different environments, right?

@Hyperkid123


Hyperkid123:
@rvsia point it to cjs(js) commonjs will work everywhere no matter what. I think we should builds for different environments eventually.

PR #278, opened 2020-01-10
Add missing tests

rvsia:
Add tests where they are missed (part one of /xxx)


rvsia:
`codecov/patch — 0% of diff hit (target 82.71%)`
 
is not applicable for tests only PR

PR #281, opened 2020-01-15
Clean obsolete props in formRenderer

rvsia:
- removes `formType`
- removes ability to parse different schemas (parsers could be exported separately)
- removes `resetAble` formType


Hyperkid123:
@rvsia snapshots

PR #286, opened 2020-01-16
Render all texts on component example

rvsia:
- inactive are hidden
- to support algolia
- there is only one header in the documentation of each component's implementation

### DIFF
--- a/packages/react-renderer-demo/src/app/pages/component-example/checkbox-multiple.md
+++ b/packages/react-renderer-demo/src/app/pages/component-example/checkbox-multiple.md
@@ -1,5 +1,3 @@
 import ComponentText from '@docs/components/component-example-text';
 
-# Checkbox multiple
-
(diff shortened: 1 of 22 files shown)


packages/react-renderer-demo/src/app/src/components/docsearch.js
```diff
@@ -52,6 +52,7 @@ const DocSearch = () => {
       indexName: 'data-driven-forms',
       inputSelector: '#data-driven-forms-search',
       handleSelected: (input, event, suggestion) => {
+        console.log(event.target.value, suggestion);
```
Hyperkid123: console log

rvsia: :man_facepalming:

rvsia:
@Hyperkid123 fixed

PR #287, opened 2020-01-17
Include query in algolia for component examples

rvsia:
It works! :dancer: :balloon: :tada: 

formssearching


Hyperkid123:
@rvsia did you try to deploy it on the test server :smile: ?

Issue #295, opened 2020-01-28
Do MUI mapper audit


rvsia:
**Scope**: MUI MAPPER

**Description**

- [componentTypes.TEXT_FIELD] :eyes: 
  - [x] tests
  - [x] validation errors
  - [x] isRequired
  - [x] isDisabled
  - [x] isReadOnly
  - [x] helperText
  - [x] description (?)
  - [x] placeholder
- [componentTypes.TEXTAREA_FIELD] :eyes:
  - [x] tests
  - [x] validation errors
  - [x] isRequired
  - [x] isDisabled
  - [x] isReadOnly
  - [x] helperText
  - [x] description (?)
  - [x] placeholder
- [componentTypes.SELECT_COMPONENT] :eyes:
  - [x] missing label :(
  - [ ] tests
  - [x] validation errors
  - [x] isRequired
  - [x] isDisabled
  - [x] isReadOnly
  - [x] helperText
  - [x] description (?)
  - [x] isClearable
  - [x] isSearchable
  - [x] isMulti (multi!)
  - [ ] loadOptions
  - [x] placeholder
- [componentTypes.CHECKBOX] :eyes:
  - [x] tests
  - [x] validation errors **error replaces the label**
  - [x] isRequired
  - [x] isDisabled
  - [x] isReadOnly
  - [x] helperText
  - [x] description (?)
- [componentTypes.CHECKBOX] multiplecheckbox  :eyes:
  - [x] tests
  - [x] validation errors
  - [x] isRequired
  - [x] isDisabled
  - [x] isReadOnly
  - [x] helperText
  - [x] description (?)
- [componentTypes.SUB_FORM] :green_heart: 
  - [x] tests
  - [x] description (?)
- [componentTypes.RADIO] :eyes:
  - [x] tests
  - [x] validation errors
  - [x] isRequired
  - [x] isDisabled
  - [x] isReadOnly
  - [x] helperText
  - [x] description (?)
- [componentTypes.TABS] :eyes:
  - [x] tests
  - [ ] description (?)
- [componentTypes.DATE_PICKER] :eyes:
  - [x] tests
  - [x] validation errors
  - [x] isRequired
  - [x] isDisabled
  - [x] isReadOnly
  - [x] helperText
  - [x] description (?)
  - [x] placeholder
  - [x] missing date format
- [componentTypes.TIME_PICKER]  :eyes:
  - [x] tests
  - [x] validation errors
  - [x] isRequired
  - [x] isDisabled
[...]

PR #296, opened 2020-01-29
MUI mapper audit fix

rvsia:
https://github.com/data-driven-forms/react-forms/issues/295

fixes https://github.com/data-driven-forms/react-forms/issues/303

Fixes formFields components

- Wizard will be done seperately
- Select still needs to be converrted into common select with loadOptions and other fancy stuff


rvsia:
@Hyperkid123 should be clear

PR #298, opened 2020-01-31
fix(pf4): Multi async select initial value change.

Hyperkid123:


packages/pf4-component-mapper/src/tests/select/select.test.js
```diff
@@ -150,6 +150,52 @@ describe('<Select />', () => {
...
+
+    setImmediate(() => {
+      wrapper.update();
+      expect(wrapper.find(Select).first().instance().state.allOptions).toEqual([{ label: 'label', value: '123' }]);
+      expect(onChange).toHaveBeenCalledWith(undefined);
+      done();
+    });
+  });
+
+  it('should load multi select Async options correctly and set initial value to ["123"]', (done) => {
```
rvsia: Could you also please add a test when values are object?

PR #302, opened 2020-02-06
Check if the field has name set, not title or key

rvsia:
- [x] check only name in validators
- [x] remove keys from all examples and included schemas
- [x] remove keys logic from all components


rvsia:
@Hyperkid123 Quick question before removing WIP: Should be it a feat or fix? It can break some forms...

Hyperkid123:
@rvsia fix

rvsia:
@Hyperkid123 done! :green_apple:

PR #308, opened 2020-02-18
feat(form-builder): split build into multiple files

Hyperkid123:


packages/react-form-renderer/rollup.config.js
```diff
@@ -29,27 +32,41 @@ function onwarn(warning) {
   throw Error(warning.message);
 }
 
-export default [{
+const plugins = [
+  nodeResolve(),
+  babel(babelOptions),
+  commonjs(commonjsOptions),
+  nodeGlobals(), // Wait for https://github.com/cssinjs/jss/pull/893
```
rvsia: what is this comment? :eyes: The mentioned PR has been merged!

packages/react-form-renderer/src/components/validator-types.js
```diff
@@ -0,0 +1,37 @@
+const validatorTypes = {
+  REQUIRED: 'required-validator',
+  /**
+     * min length of the input value
```
rvsia: what about removing these comments? Sounds pretty redundant to me.

packages/react-form-renderer/rollup.config.js
Hyperkid123: Oh right this was due to SSR. We should keep that one. I will just remove the comment

Hyperkid123:
@rvsia renamed the commit

PR #312, opened 2020-02-19
fix(pf4): scroll to top on switching to a different step

boaz0:
For instance, when there are two step that are too long for the wizard container (e.g. too many fields) and are next to each other. If you scroll down in one of them and switch to the other, the
location of your scroll bar won't reset to the top but will stay at the bottom.

This is bad because the user expects the scroll bar to go back to the top of the next step.

//cc @rvsia

### DIFF
--- a/packages/pf4-component-mapper/src/form-fields/wizard/wizard-step.js
+++ b/packages/pf4-component-mapper/src/form-fields/wizard/wizard-step.js
@@ -4,38 +4,59 @@ import { Title } from '@patternfly/react-core/dist/js/components/Title/Title';
 import PropTypes from 'prop-types';
 import WizardStepButtons from './step-buttons';
 
-export const RenderTitle = ({ title, customTitle }) => customTitle ? customTitle : <Title headingLevel="h1" size="xl">{ title }</Title>;
+export const RenderTitle = ({ title, customTitle }) => customTitle ? customTitle : <Title headingLevel="h1" size="xl">{title}</Title>;
 
 RenderTitle.propTypes = {
   title: PropTypes.node,
   customTitle: PropTypes.node,
 };
 
-const WizardStep = ({
-  title,
-  description,
-  fields,
-  formOptions,
-  showTitles,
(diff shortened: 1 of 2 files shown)


packages/pf4-component-mapper/src/form-fields/wizard/wizard-step.js
```diff
@@ -4,38 +4,61 @@ import { Title } from '@patternfly/react-core/dist/js/components/Title/Title';
...
+    // to fill all the data for the next step. If the next step contains instructions
+    // at the top, the user will miss them because the scrollbar offset will stay at
+    // the same place it was.
+    if(prevProps.stepKey !== this.props.stepKey) {
+      // HACK: I can not pass ref to WizardBody because it is not
+      // wrapped by forwardRef. However, the step body (the one that overflows)
+      // is the grand parent of the form element.
+      const stepBody = this.formRef.current && this.formRef.current.parentNode && this.formRef.current.parentNode.parentNode;
+      if(stepBody && stepBody.scrollTo && typeof stepBody.scrollTo === 'function') {
+        stepBody.scrollTo({top: 0, left: 0, behavior: 'smooth'});
```
boaz0: I mean.... I don't know why on enzyme scrollTo is not a function but I have to run this check to make tests pass.

Hyperkid123: Enzyme does not recognize s roll because the test env runs on Node and it does not have browser functions and prototypes like Element or scroll. These are only browser specific. To fix this you should add a scroll to polyfill to test setup.

Hyperkid123: If you want to know what functions are missing in test env, I would recommend looking at jasmíne and JSDom, because that is what is jest using to simulate DOM in node.

boaz0:
Thanks a lot @Hyperkid123 & @rvsia 
That was a great experience as a contributor. :grinning:

PR #316, opened 2020-02-20
Upgrade final form

Hyperkid123:
- updates react final form
- temporarily disabled multiple-choice list (buggy)

PR #317, opened 2020-02-21
[V2] Use form api instead of formOptions

rvsia:
Usage

```jsx
import { useFormApi } from '@data-driven-forms/react-form-renderer';

const Component = (props) => {
     const formOptions = useFormApi();
     ...
}
```

TODO:
- [x] export context as well
- [ ] update FieldProvider after https://github.com/data-driven-forms/react-forms/pull/316 is merged
**FieldProvider is temporarily wrapped in a function, so it would be easier to rebase**

- [x] convert Wizard components to functions
  - MUI wizard is just wrapped, it used only in demo, it needs full implementation
  - PF3 wizard converted
  - PF4 wizard still class, useFormApi is used in the wrapper (huge task to refactor it, will do it in separate PR)
- [x] FREAKIN' TESTS :bomb: :boom: :boom:

### DIFF
--- a/__mocks__/mock-form-template.js
+++ b/__mocks__/mock-form-template.js
@@ -1,6 +1,9 @@
 import React from 'react';
+import { useFormApi } from '@data-driven-forms/react-form-renderer';
 
-const FormTemplate = ({ schema: { title, label, description }, formFields, formOptions, FormSpy }) => {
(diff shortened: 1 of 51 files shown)


rvsia:
```jsx
import { formTemplate } from ‘@data-driven-forms/pf4-component-mapper’

<FormRenderer {...} formTemplate={formTemplate({showFormControls: false)} />
```

PR #322, opened 2020-02-24
Export field array instead of using a prop

rvsia:
```jsx
import { FieldArray } from '@data-driven-forms/react-form-renderer';
```


Hyperkid123:
> ```js-jsx
> import { FormSpy } from '@data-driven-forms/react-form-renderer';
> ```

FieldArray. :scissors:

PR #327, opened 2020-02-25
[v2]: Export field provider

Hyperkid123:
### Changes
- fields are no longer automatically wrapped in field provider
- mapper now have to use either `useFielProviderApi` hook or `FieldProvider` directly from `react-form-renderer` library
- PF4 Field array should be reworked since this is a major change and out of the scope of this PR. Tests have been disabled
- some tests have been disabled due to no longer existing APIs. That should be replaced in a future


packages/react-form-renderer/src/components/field-provider.js
```diff
@@ -1,72 +1,18 @@
...
-import { useField } from 'react-final-form';
+import React from 'react';
 import PropTypes from 'prop-types';
 
-import enhancedOnChange from '../form-renderer/enhanced-on-change';
 import dataTypes from './data-types';
-import useFormApi from '../hooks/useFormApi';
+import useFieldProviderApi from '../hooks/use-field-provider-api';
 
-const FieldProvider = ({ name, initializeOnMount, component, render, ...props }) => {
```
rvsia: You can remove most of propTypes here

packages/react-form-renderer/src/hooks/use-field-provider-api.js
```diff
@@ -0,0 +1,59 @@
...
+   * Prepare deleted value of field
+   */
+  const fieldClearedValue = props.hasOwnProperty('clearedValue') ? props.clearedValue : formOptions.clearedValue;
+
+  /**
+   * construct component props necessary that would live in field provider
+   */
+  return {
+    ...props,
+    formOptions,
```
rvsia: Do we really want to pass formOptions here? I thought that was the reason we made the separate hook.

packages/react-form-renderer/src/index.js
```diff
@@ -13,6 +13,7 @@ export { default as RendererContext } from './components/renderer-context';
 export { default as FormSpy } from './components/form-spy';
 export { default as Form } from './components/form';
 export { default as FieldArray } from './components/field-array';
+export { default as useFieldProviderApi } from './hooks/use-field-provider-api';
```
rvsia: What about the name? I think it's too long and for other people who doesn't know implementation, it can be unnecessary specific. (_What is a provider?_)

I think `useFieldApi` (and it will be consistent with `useFormApi`) sounds better (as it is basically just wrapped useField from FF)

rvsia:
The failing test has been removed in #326.

packages/react-form-renderer/src/hooks/use-field-provider-api.js
Hyperkid123: I want to make props clean up in the next step. There is a lot more stuff like that.

rvsia: I have already cleaned up formOptions in the PR with that hook, so I am think it should be ready to be removed now.

Hyperkid123:
@rvsia field array is probably not working and wizards might have some missing FieldProviders

rvsia:
_Notes (note related do changes in this PR): we need to solve isMulti / multi in the future_


**MATERIAL UI MAPPER**

MUI switch on/off text does not work

Probably because

```jsx
label={<FormLabel>{input.value ? onText || label : offText || label}</FormLabel>}
```

maybe because input.value is no longer valid, shouldn't it be `checked` or something like that?

rvsia:
_Notes (not related do changes in this PR): not sure, but isReadOnly select is not readOnly_


**PF3 MAPPER**

Looks OK :ok:

rvsia:
_Notes (not related do changes in this PR): checkbox is required does not have the asterisk_


**PF3 MAPPER**

Looks OK :ok:

rvsia:
@Hyperkid123 I have finished the UI testing! :) Looks great, just the MUI switch needs more attention. (But maybe it's not related to this PR at all....)

PR #339, opened 2020-02-28
[V2] Validator mapper

rvsia:
Allows to use a custom validator mapper

```jsx
const customValidatorMapper = {
  custom: () => (value) => value > 6 ? 'Value is bigger than 6' : undefined
}

const schema = {
  fields: [{
   name: 'name',
   component: 'text-field',
   validate: [{type: 'custom'}]
  }]
}

<FormRenderer 
  ...
  schema={schema}
  validatorMapper={customValidatorMapper}
/>
```

- This mapper is merged with the default one (user can rewrite the default ones)
- schemaValidator still controls if the type is in validatorMapper (even the custom ones!)

(and as a bonus, empty validation is no longer passed as a prop!)


rvsia:
@Hyperkid123 Yeah, we should export most of inner functions we are using to create the default validator mapper and write about them.

PR #342, opened 2020-03-02
[V2] Action mapper proposal

rvsia:
```jsx

// Returns a function
const actionMapper = {
    loadData: (data) => () => new Promise((resolve) => setTimeout(() => resolve({ custom: 'ererewr', ...data }), 1700))
}

const schema = {
  title: 'Testing dialog',
  description: 'Description of testing Dialog',
  fields: [
    {
      name: 'dataShower',
      component: 'dataShower',
      // default
      loadOptions: () => new Promise((resolve) => setTimeout(() => resolve({ custom: 'data' }), 1500))
    },
    {
      name: 'dataShower',
      component: 'dataShower',
      // use of actions
      actions: {
        loadOptions: ['loadData', {x: 'y'}]
      }
    }
  ]
};

const App = () => (
    <div>
        <FormRenderer
            ...
            schema={schema}
            actionMapper={actionMapper}
        />
    </div>
)
```

@Hyperkid123 What do you think?

### DIFF
--- a/packages/react-form-renderer/demo/sandbox.js
+++ b/packages/react-form-renderer/demo/sandbox.js
@@ -105,6 +105,21 @@ const output = {
       title: 'Text Box string value',
       dataType: 'string',
       component: componentTypes.TEXT_FIELD
+    },
+    {
+      name: 'dataShower',
+      label: 'Test select',
+      component: 'dataShower',
+      loadOptions: () => new Promise((resolve) => setTimeout(() => resolve({ custom: 'data' }), 1500))
+    },
+    {
+      name: 'dataShower',
+      label: 'Test select',
+      component: 'dataShower',
+      actions: {
+        loadOptions: ['loadData', { x: 'y', pp: 'adsad' }],
+        label: ['loadLabel', 'cosi']
+      }
     }
   ]
 };
--- a/packages/react-form-renderer/src/components/form-renderer.js
+++ b/packages/react-form-renderer/src/components/form-renderer.js
@@ -22,7 +22,8 @@ const FormRenderer = ({
   subscription,
   clearedValue,
   schema,
-  validatorMapper
+  validatorMapper,
+  actionMapper
 }) => {
   let schemaError;
 
@@ -30,7 +31,8 @@ const FormRenderer = ({
 
   try {
     const validatorTypes = Object.keys(validatorMapperMerged);
-    defaultSchemaValidator(schema, formFieldsMapper, validatorTypes);
+    const actionTypes = actionMapper ? Object.keys(actionMapper) : [];
+    defaultSchemaValidator(schema, formFieldsMapper, validatorTypes, actionTypes);
   } catch (error) {
     schemaError = error;
     console.error(error);
@@ -58,6 +60,7 @@ const FormRenderer = ({
           value={{
             formFieldsMapper,
             validatorMapper: validatorMapperMerged,
+            actionMapper,
             formOptions: {
               pristine,
               onSubmit,
@@ -94,13 +97,14 @@ FormRenderer.propTypes = {
   subscription: PropTypes.shape({ [PropTypes.string]: PropTypes.bool }),
   clearedValue: PropTypes.any,
   formFieldsMapper: PropTypes.shape({
-    [PropTypes.string]: PropTypes.oneOfType([PropTypes.node, PropTypes.element, PropTypes.func]).isRequired
+    [PropTypes.string]: PropTypes.oneOfType([PropTypes.node, PropTypes.element, PropTypes.func])
   }).isRequired,
-  formTemplate: PropTypes.shape({
-    [PropTypes.string]: PropTypes.oneOfType([PropTypes.node, PropTypes.element, PropTypes.func]).isRequired
-  }).isRequired,
+  formTemplate: PropTypes.oneOfType([PropTypes.node, PropTypes.element, PropTypes.func]).isRequired,
   validatorMapper: PropTypes.shape({
-    [PropTypes.string]: PropTypes.func.isRequired
+    [PropTypes.string]: PropTypes.func
+  }),
+  actionMapper: PropTypes.shape({
+    [PropTypes.string]: PropTypes.func
   })
 };
 
--- a/packages/react-form-renderer/src/hooks/use-field-api.js
+++ b/packages/react-form-renderer/src/hooks/use-field-api.js
@@ -1,9 +1,11 @@
-import { useEffect } from 'react';
+import { useEffect, useContext } from 'react';
 import { useField } from 'react-final-form';
 import useFormApi from './use-form-api';
 import enhancedOnChange from '../form-renderer/enhanced-on-change';
+import RendererContext from '../components/renderer-context';
 
 const useFieldApi = ({ name, initializeOnMount, component, render, validate, dataType, ...props }) => {
+  const { actionMapper } = useContext(RendererContext);
   const formOptions = useFormApi();
   const fieldProps = useField(name, { validate, ...props });
 
@@ -33,10 +35,22 @@ const useFieldApi = ({ name, initializeOnMount, component, render, validate, dat
   const fieldClearedValue = Object.prototype.hasOwnProperty.call(props, 'clearedValue') ? props.clearedValue : formOptions.clearedValue;
 
   /**
+   * Map actions to props
+   */
+  let overrideProps = {};
+  if (props.actions) {
+    Object.keys(props.actions).forEach((prop) => {
+      const [action, ...args] = props.actions[prop];
+      overrideProps[prop] = actionMapper[action](...args);
+    });
+  }
+
+  /**
    * construct component props necessary that would live in field provider
    */
   return {
     ...props,
+    ...overrideProps,
     ...fieldProps,
     input: {
       ...fieldProps.input,
--- a/packages/react-form-renderer/demo/form-fields-mapper.js
+++ b/packages/react-form-renderer/demo/form-fields-mapper.js
@@ -42,6 +42,33 @@ const SelectField = (props) => {
   );
 };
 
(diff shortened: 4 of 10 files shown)


packages/react-form-renderer/demo/sandbox.js
```diff
@@ -1,119 +1,26 @@
...
-        {
-          when: 'text_box_3',
-          is: 'x'
+      name: 'dataShower',
+      label: 'Test select',
+      component: 'dataShower',
+      actions: {
+        loadOptions: {
+          action: 'loadData',
+          data: { x: 'y' }
```
Hyperkid123: I would call this attribute `args` or `arguments` to make it clear, that it is a function argument. Data might seem misleading based on the action type (function/promise)

packages/react-form-renderer/src/hooks/use-field-api.js
```diff
@@ -32,11 +34,20 @@ const useFieldApi = ({ name, initializeOnMount, component, render, ...props }) =
    */
   const fieldClearedValue = Object.prototype.hasOwnProperty.call(props, 'clearedValue') ? props.clearedValue : formOptions.clearedValue;
 
+  let overrideProps = {};
+  if (props.actions) {
+    Object.keys(props.actions).forEach((prop) => {
```
Hyperkid123: There is a question of `undefined` values here. What if the actions is this:
```jsx
{
  someAction: undefined
}
```

Do we want to override the existing action with the `undefined` value and possibly risk runtime error or do we want to preserve the original prop?

Maybe we can make it more specific and if the user want to override the prop with an empty one, we can enforce ussage of `false` or `null` or something similar?

packages/react-form-renderer/demo/sandbox.js
rvsia: No, the action will receive the whole object, data is just a name of custom attribute

packages/react-form-renderer/src/hooks/use-field-api.js
rvsia: What about to enforce this in the schema validator? No undefined values, etc.

packages/react-form-renderer/demo/sandbox.js
rvsia:
```jsx
action: 'loadData',
superCustomData: { x: 'y' }
```

``jsx
action: 'loadData',
url: 'neco.com'
```

etc.

Comment (Hyperkid123) [reply packages/react-form-renderer/demo/sandbox.js]: Ah, I thought that it will only accept the data argument. In that, I would go with a slightly different syntax, so we don't force object arguments on functions and people won't have to refactor existing functions:
```jsx
{
  actions: {
    loadOptions: ['loadData' /*name*/, ['foo', 'bar', {x: 'y'}] /*function args*/]
  }
}

// usage
const [name, args] = whereEverTheStuffComesFrom()
loadOptions(...args)
```

packages/react-form-renderer/src/hooks/use-field-api.js
Hyperkid123: Well, the question is more about if we should do it or not. I can imagine if somebody is generating this, thew will have `undefined` keys instead of deleting them. Maybe use the `false`/`null` to override the prop to be empty?

packages/react-form-renderer/demo/sandbox.js
rvsia: No, users don't have to refactor anything.

The action mapper looks like this

`loadData: ({ url, query }) => () => origFunc(url, query)`

And I would keep the object notation, as it is consistent with the way how validators are written and also it is definitely more readable than an array.

packages/react-form-renderer/src/hooks/use-field-api.js
rvsia: But why, if you want to override props in the schema, just don't write the original ones... :D  and if someone generates this, so they will just write their generator without generating falsy values.

This should serve to map actions to props, not to override them...

packages/react-form-renderer/demo/sandbox.js
rvsia: And also, there is no need to do anything in the component itself, function is just replaced by function, as the mapper returns function.

Hyperkid123: I think you missed my point. Forget the load options in the example. What if somebody is using classic `loadOptions(url, query)` arguments they would have create a wrapper around that function to accept arguments form the action mapper. I would like to avoid this:
```jsx
loadOptionsWrapper({url, query}) => loadOptions(url, query)
```

By using the object, we force the object on the action arguments themselves. If you use the spread operator on the argument array, you don't have to do a thing. If you want an object just use an object in the argument array. Keep in mind this action mapper affects every component, not just the select. You can potentially use this to override every single prop of any component.

packages/react-form-renderer/src/hooks/use-field-api.js
Hyperkid123: Yup that is true.

packages/react-form-renderer/demo/sandbox.js
rvsia: I don't think that creating a wrapper is so huge problem, the advantages that the object notation brings are much more valuable. (And as this is a new feature, there is no refactoring at all...)

Let's say we have an action for translating labels:

Variant a) Object
```jsx
const actionMapper = {
    loadLabel: ({ id }) => intl(id)
}
```

```jsx
{
  name: 'test',
  component: 'text-field',
  actions: {
    label: {
      action: 'loadLabel',
      id: 'cosi'
     }
   }
}
```

With no context, you already know what it does.

Variant b) Array
```jsx
const actionMapper = {
    loadLabel: intl
}
```

It's not clear what the action does, no interface.

```jsx
{
  name: 'test',
  component: 'text-fields',
  actions: {
    label: [ 'loadLabel', 'cosi' ]
   }
}
```

- It's not clear what the array means, what the `cosi` is and with more arguments you need to **keep the order**... 

- No args:

```jsx
label: ['loadLabel']
// or ?
label: 'loadLabel'
```

Then, you have two ways how to do it, or you won't support using just a string and users will be confused as they know this notation from webpack and other configs.

- or let's say we need to skip the first three arguments:

```jsx
label: ['loadLabel', undefined, undefined, undefined, 'cosi']
```

and if you store these values in the database and actions on frontend, then you basically lost all information you need to be able to successfully create the action schema.

Another advantage of the object is that you can validate its schema easily (_are you using loadLabel without id?!_)

Of course, users can simulate the object with passing of object (and they will do as using array is not intuitive), but they can also simulate args array with the same way, if they need:

```
 loadLabel: ({ args }) => intl(...args)

label: {
   action: 'loadLabel',
   args: ['cosi']
}
```

And as users will use object anyway, I think it's better to support good practices. Enforcing is not always bad thing.

Hyperkid123: > I don't think that creating a wrapper is so huge problem, the advantages that the object notation brings are much more valuable

Advantages aside (i do agree here), you are still trying to enforce certain notation of functions. Js functions take an array of arguments and not object.

> And as users will use object anyway, I think it's better to support good practices.

We are not making a linter library. We are not here to decide what is a good/bad practice. If somebody wants to have 10 `undefined` parameters before passing anything valuable, it is up to them. The only thing that we should "enforce" is JS API. And as I said before, functions take an array as an argument, not objects.

Imagine someone crazy that wants to use autogenerated API from OpenAPI. It will generate your functions where you can easily have a ton of undefined values and you have 150 of different callbacks you use in your schemas. Do you really want to make a wrapper around every single one of them?

> It's not clear what the array means, what the cosi is and with more arguments you need to keep the order...

Again I agree, but again you have different tools for adding meaning to attributes like this (JSDoc).

rvsia: > We are not here to decide what is a good/bad practice.

Actually, we are. It's up to us how we define the schema. (Just take a look on validators/ conditions, their notation is basically the same as I propose here. And JavaScript is just current implementation, the schema should not be adjusted because of it.)

> 150 of different callbacks you use in your schemas. Do you really want to make a wrapper around every single one of them?

One-line function. Not great, not terrible.

> (JSDoc).

Using JSDoc (or any of other types of comments) to tell what the code does is not very good idea. Code code should be self-explanatory. (Especially schemas, that can be stored without the code.)

---

But yeah, I will change it.

packages/react-form-renderer/src/components/form-renderer.js
```diff
@@ -22,15 +22,17 @@ const FormRenderer = ({
...
+  actionMapper
 }) => {
   let schemaError;
 
   const validatorMapperMerged = { ...defaultValidatorMapper, ...validatorMapper };
 
   try {
     const validatorTypes = Object.keys(validatorMapperMerged);
-    defaultSchemaValidator(schema, formFieldsMapper, validatorTypes);
+    const actionTypes = actionMapper ? Object.keys(actionMapper) : {};
```
Hyperkid123: ` Object.keys(actionMapper)` will return `true` even on an empty object. Is that desired?

Hyperkid123: Furthermore, sometimes the `actionTypes` is an array and sometimes object.

rvsia: Yeah, it won't do any harm.

rvsia: > Furthermore, sometimes the actionTypes is an array and sometimes object.

What you think? actionTypes should be array with names of the actions used to validate the schema

Hyperkid123: SO maybe have the fallback value `[]` instead of `{}`?

rvsia: oh :hankey:  you are right :D

PR #348, opened 2020-03-03
[V2] Fix multiple checkboxes

rvsia:


Hyperkid123:
:skull_and_crossbones:

packages/common/src/multiple-choice-list.js
```diff
@@ -1,70 +1,52 @@
...
+      {options.map((option) => (
+        <SingleCheckbox
+          Checkbox={Checkbox}
+          aria-label={option['aria-label'] || option.label}
+          {...rest}
+          value={option.value}
+          label={option.label}
+          name={input.name}
+          option={option}
+          id={`${rest.id || input.name}-${option.value}`}
```
Hyperkid123: The `rest.id` won't work on its own. All checkboxes will have the same id. Either add the -option.value or we should remove the id entirely and just use name/option combination.

rvsia: there is a `-${option.value}`

PR #350, opened 2020-03-03
[V2] breaking(renderer): remove component types with -field affix.

Hyperkid123:


Hyperkid123:
demo died

rvsia:
Please do not forget to update the migration guide! :bookmark: :book:

Hyperkid123:
I think the demo might be broken until we release a new version of the renderer

PR #352, opened 2020-03-04
Rewrite mappers FormTemplates to be components.

Hyperkid123:
### Changes
- rewrite mappers FormTemplates to be components.
- onStateUpdate is back as a FormRenderer prop

### DIFF
--- a/packages/react-form-renderer/src/components/form-renderer.js
+++ b/packages/react-form-renderer/src/components/form-renderer.js
@@ -9,10 +9,28 @@ import renderForm from '../form-renderer/render-form';
 import defaultSchemaValidator from '../parsers/default-schema-validator';
 import SchemaErrorComponent from '../form-renderer/schema-error-component';
 import defaultValidatorMapper from '../validators/validator-mapper';
+import useComponentSpy from '../hooks/use-component-spy';
+
+const FormContent = ({ value, formFields, schema, onStateUpdate, Template }) => {
+  useComponentSpy(onStateUpdate);
+  return (
+    <RendererContext.Provider value={value}>
+      <Template formFields={formFields} schema={schema} />
+    </RendererContext.Provider>
+  );
+};
+
+FormContent.propTypes = {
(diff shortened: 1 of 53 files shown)


packages/react-form-renderer/src/components/form-renderer.js
```diff
@@ -55,38 +82,43 @@ const FormRenderer = ({
...
-              renderForm,
-              ...mutators,
-              ...form
-            }
-          }}
-        >
-          <FormTemplate formFields={renderForm(schema.fields)} schema={schema} />
-        </RendererContext.Provider>
-      )}
+      render={({ handleSubmit, pristine, valid, form: { reset, mutators, getState, submit, ...form }, ...state }) => {
```
rvsia: Unnecessary return block

packages/react-form-renderer/src/components/form-renderer.js
```diff
@@ -10,9 +10,35 @@ import defaultSchemaValidator from '../parsers/default-schema-validator';
 import SchemaErrorComponent from '../form-renderer/schema-error-component';
 import defaultValidatorMapper from '../validators/validator-mapper';
 
+const useComponentSpy = (onStateUpdate) => {
```
rvsia: What about to move this to hooks folder?

Hyperkid123: I could. If you reaaaaaally want me to

rvsia: Yes, please :D

packages/react-form-renderer/src/tests/form-renderer/render-form.test.js
```diff
@@ -1185,7 +1185,7 @@ describe('renderForm function', () => {
     });
   });
 
-  it('should use actionMapper', () => {
+  it.only('should use actionMapper', () => {
```
rvsia: only ?

PR #359, opened 2020-03-06
feat(pf4): add duallist component

### LINKED ISSUE #358: dual-list-select is not available in PF4
**Scope**: PF4 dual-list-select

**Description**
Attempted to use `dual-list-select` in a PF4 app and hit the following error:
dual-list-select

**Schema**

Use the example from here:
https://data-driven-forms.org/others/miq-components#duallistselect

rvsia:
closes #358 

Add Dual List 'placeholder' component

image

duallistselect

@chambridge  @Hyperkid123 

What do you think about the design?
Will you also need async filtering?

todo:
- [x] docs
- [x] tests
- [x] ~~async filtering for options~~ (in the future)

### DIFF
--- a/packages/pf4-component-mapper/src/form-fields/dual-list-select.scss
+++ b/packages/pf4-component-mapper/src/form-fields/dual-list-select.scss
@@ -1,0 +1,12 @@
+.ddorg__pf4-component-mapper__dual-list-select-toolbar {
+    height: auto !important;
+}
+
+.ddorg__pf4-component-mapper__dual-list-select {
+    height: 300px !important;
+    background: none !important;
+}
+
+.ddorg__pf4-component-mapper__dual-list-select-option {
+    height: 100% !important;
+}
--- a/packages/pf4-component-mapper/demo/index.html
+++ b/packages/pf4-component-mapper/demo/index.html
@@ -4,7 +4,7 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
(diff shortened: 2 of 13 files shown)


chambridge:
@rvsia Looks good to me.

Can you clarify async filtering? are you meaning in order to make http requests on the dataset? That is a nice to have, but I think we can make it work if we don't have that.

I think what you show above would meet our needs.

rvsia:
@chambridge Yeah, I am talking about filtering using API's calls. Our API has a maximum limit of 100 returned items, so I can imagine that if users has more than 100 OpenShift labels, it would be problem to show them all without asynchronous filtering.

chambridge:
That would certainly be nice, but since the API is building the data driven forms it in theory it could collect them (there is just a potential performance impact) to collect all the labels when settings is called.

Hyperkid123:
@rvsia we can add async filtering later. We might find more requirements after this component is published and used.

rvsia:
@chambridge Do you want to test this PR too?

packages/pf4-component-mapper/src/form-fields/dual-list-select.scss
```diff
@@ -0,0 +1,18 @@
+.ddorg__pf4-component-mapper__dual-list-select-toolbar {
+    height: auto !important;
+    padding: 0px;
+}
+
+.ddorg__pf4-component-mapper__dual-list-select-toolbar-content {
+    padding: 0px 0px 16px 0px !important;
```
Hyperkid123: you can us pf modifier classes like `pf-p-m-0` `pf-u-pr-md` etc. So we can remove the !important nonsense :smile:

rvsia: let's them make the nonsense :see_no_evil:

rvsia:
> @rvsia can you add the addons (just the classes we used) to the docs as well?

They are already there

PR #364, opened 2020-03-09
feat(renderer): Allow defining renderer props in form schema.

### LINKED ISSUE #363: Allow to pass FormRenderer props in the top level object of the schema
```jsx
{clearedValue: null, fields: [{ .... }]}
```

Hyperkid123:
closes #363

### DIFF
--- a/packages/react-form-renderer/src/components/form-renderer.js
+++ b/packages/react-form-renderer/src/components/form-renderer.js
@@ -45,6 +45,12 @@ const FormRenderer = ({
   onStateUpdate
 }) => {
   let schemaError;
+  const {
+    initialValues: schemaInitialValues,
+    clearOnUnmount: schemaClearOnUnmount,
+    subscription: schemaSubscription,
+    clearedValue: schemaClearedValue
+  } = schema;
 
   const validatorMapperMerged = { ...defaultValidatorMapper, ...validatorMapper };
 
@@ -71,9 +77,9 @@ const FormRenderer = ({
       onSubmit={onSubmit}
       mutators={{ ...arrayMutators }}
       decorators={[createFocusDecorator()]}
-      initialValues={initialValues}
+      initialValues={schemaInitialValues || initialValues}
       validate={validate}
-      subscription={{ pristine: true, submitting: true, valid: true, ...subscription }}
+      subscription={{ pristine: true, submitting: true, valid: true, ...subscription, ...schemaSubscription }}
       render={({ handleSubmit, pristine, valid, form: { reset, mutators, getState, submit, ...form }, ...state }) => (
         <FormContent
           formFields={renderForm(schema.fields)}
@@ -91,11 +97,11 @@ const FormRenderer = ({
               onReset,
               getState,
               valid,
-              clearedValue,
+              clearedValue: typeof schemaClearedValue !== 'undefined' ? schemaClearedValue : clearedValue,
               submit,
               handleSubmit,
               reset,
-              clearOnUnmount,
+              clearOnUnmount: typeof schemaClearOnUnmount !== 'undefined' ? schemaClearOnUnmount : clearOnUnmount,
               renderForm,
               ...mutators,
               ...form
--- a/packages/react-renderer-demo/src/app/pages/renderer/renderer-api.md
+++ b/packages/react-renderer-demo/src/app/pages/renderer/renderer-api.md
@@ -1,11 +1,57 @@
 import RouterLink from 'next/link';
 import Link from '@material-ui/core/Link';
-import Grid from '@material-ui/core/Grid'
+import Grid from '@material-ui/core/Grid';
+import Card from '@material-ui/core/Card';
+import CardContent from '@material-ui/core/CardContent';
+import AnnouncementIcon from '@material-ui/icons/Announcement'
+import orange from '@material-ui/core/colors/orange'
+import Typography from '@material-ui/core/Typography';
+import Box from '@material-ui/core/Box';
 
 import ListOfContents from '../../src/helpers/list-of-contents';
 
 <Grid container item>
 <Grid item xs={12} md={10}>
+
+
+# Schema
+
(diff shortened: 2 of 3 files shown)


packages/react-renderer-demo/src/app/pages/renderer/renderer-api.md
```diff
@@ -1,12 +1,58 @@
...
+|description|node|<RouterLink href="/renderer/component-mapping#description"><Link href="/renderer/component-mapping#description">Description</Link></RouterLink> of the form. Optional.|
+|fields|array of objects|<RouterLink href="/renderer/component-api"><Link href="/renderer/component-api">Components</Link></RouterLink> of the form. Required!|
+|initialValues|object|An object of fields names as keys and values as their values.|
+|<RouterLink href="/renderer/unmounting"><Link href="/renderer/unmounting">clearOnUnmount</Link></RouterLink>|bool|Will clear values of unmounted components. You can also set this to specific component in the form schema.|false|
+|subscription|object|You can pass your own [subscription](https://final-form.org/docs/react-final-form/types/FormProps#subscription), which will be added to default settings.|`{ pristine: true, submitting: true, valid: true }`|
+|<RouterLink href="/renderer/cleared-value"><Link>clearedValue</Link></RouterLink>|any|Value that will be set to field with **initialValue** after deleting it. Useful for forms while editing.|undefined|
```
rvsia: without ?

packages/react-renderer-demo/src/app/pages/renderer/renderer-api.md
```diff
@@ -1,12 +1,58 @@
...
+import Typography from '@material-ui/core/Typography';
+import Box from '@material-ui/core/Box';
 
 import ListOfContents from '../../src/helpers/list-of-contents';
 
 <Grid container item>
 <Grid item xs={12} md={10}>
 
+
+# Schema
```
rvsia: Why it is moved above the renderer-api props?

Hyperkid123: Because I think it was easily missed before.

packages/react-renderer-demo/src/app/pages/renderer/renderer-api.md
Hyperkid123: :eyes: where i can't find it

rvsia: Oh, sorry, it's okay.

Maybe I would change 

Value that will be set to field with **initialValue** after deleting it

to 

Value that will be set to fields with **initialValue** when user clears the field.

packages/react-form-renderer/src/components/form-renderer.js
```diff
@@ -91,11 +97,11 @@ const FormRenderer = ({
               onReset,
               getState,
               valid,
-              clearedValue,
+              clearedValue: typeof schemaClearedValue !== 'undefined' ? schemaClearedValue : clearedValue,
```
rvsia: Shouldn't it check if the property exists? What if user wants to use `undefined` as the clearedValue because the renderer component is using something else?

packages/react-form-renderer/src/components/form-renderer.js
```diff
@@ -91,11 +97,11 @@ const FormRenderer = ({
...
               submit,
               handleSubmit,
               reset,
-              clearOnUnmount,
+              clearOnUnmount: typeof schemaClearOnUnmount !== 'undefined' ? schemaClearOnUnmount : clearOnUnmount,
```
rvsia: same as above ^^

packages/react-form-renderer/src/components/form-renderer.js
Hyperkid123: undefined is set by default

rvsia:
```jsx
schema = {
  ....
  clearedValue: undefined,
  fields: ....
}

<FormRenderer ... clearedValue={null} />
```

Now, we are breaking the priority rule. (Yeah, the use case does not make much sense, but everything can happen.)

Hyperkid123: That brings up the question again. Should we just move these props exclusively to the schema? And remove them from renderer props?

rvsia: These props are more implementation details stuff, so I would rather keep them outside schemas. At least some of them - having subscription defined in the schema does not make much sense to me, as it would be in each form the same. And also we should define some kind of boundary, what should be in the schema and what not. For example, we could also define onSubmit, onCancel in the schema, but that would mean a lot of function passing... so it's a tough question. :D

Hyperkid123: I think that initial values, cleared value, and clearOnUmnout should be in the schema and the rest should be props. I would like to avoid having duplications though.

rvsia: clearedValue and clearOnUnmount; I agree. clearedValue is basically useful mostly for edit forms, so it makes sense to have it only in the schema. clearOnUnmnout: I haven't seen any use case yet, so, who knows :D ... however, I see an issue with initialValues as they are loaded outside the form, so that mean you will do a lot of:

```jsx
<Renderer
   ...
   schema={schema(initialValues)}
/>
```

Hyperkid123: You are right

Hyperkid123:
@rvsia the more I think about this, the more I think this is a mistake. FOr example, cleared value.

If I want to have the same schema for creating/editing I would now have to modify the schema (one or another) if I want to add `clearedValue`. Also, if we share the schema between multiple apps and each has a different requirement for cleared value, you would have to again modify each schema, instead of having different props in your code. I think that this might not be such a good idea after all.

rvsia:
@Hyperkid123 Yeah, the initial thought was to have it as an alternative way, that wouldn't break anything. With introducing of Action mapper/Validation mapper, we are getting close to use schemas as the only source of info for forms, so I was thinking about moving more into this direction. But since we are trying to remove all multiple ways of controlling the form, I think we can leave this idea alone and close it.

PR #371, opened 2020-03-10
[V2] Update doc

rvsia:
Updates DOCS for changes introduced in V2.

All new features are documented, except the bundle size optimization imports.

There is also a migration guide page with the changed stuff.


packages/react-renderer-demo/src/app/pages/renderer/get-started.md
```diff
@@ -9,9 +9,7 @@ import RawComponent from '@docs/raw-component';
 
 # React form renderer
 
-React form renderer is a component designed for ManageIQ and
-Insights projects that takes JSON form definitions
-and renders them into react components.
+Data Driven Forms converts JSON form definitions into fully functional React forms.
 It uses [React Final Form](https://github.com/final-form/react-final-form) for the form state management.
 It is highly recommended to check their documentations first to fully understand how
 the [Data Driven Dorms](https://github.com/data-driven-forms/react-forms) libraries work.
```
Hyperkid123: Can you fix this typo please?

packages/react-renderer-demo/src/app/pages/renderer/component-mapping.md
```diff
@@ -7,26 +7,16 @@ import ListOfContents from '../../src/helpers/list-of-contents';
 <Grid container item>
 <Grid item xs={12} md={10}>
 
-# Introduction
+# ComponentMapper
```
Hyperkid123: There are `title` and `description` attributes in schema that are no longer rendered automatically. Can you move these to the form template example?

packages/react-renderer-demo/src/app/pages/renderer/component-api.md
```diff
@@ -184,8 +184,8 @@ Wizard step <br/>
 |Prop|Type|Description|
```
Hyperkid123: There are some legacy comments about form groups `/renderer/component-api#formgroupwrappedcomponents`. I would probably remove this. I would probably change this to something like supported mappers prop names or something. To let people know that if they want their mappers compatible with the ones living in data-driven forms, they should use this notation.

packages/react-renderer-demo/src/app/pages/renderer/form-template.md
```diff
@@ -0,0 +1,70 @@
...
+import Link from '@material-ui/core/Link';
+
+import ListOfContents from '../../src/helpers/list-of-contents';
+
+<Grid container item>
+<Grid item xs={12} md={10}>
+
+# FormTemplate
+
+FormTemplate component allows you to fully customize the form. FormTemplate receives only two props:
```
Hyperkid123: I would probably add the `layout` word there

FormTemplate component allows you to fully customize the form **layout**. FormTemplate receives only two props

packages/react-renderer-demo/src/app/pages/renderer/form-template.md
```diff
@@ -0,0 +1,70 @@
...
+
+<FormRenderer
+  ...
+  FormTemplate={(props) => <FormTemplate {props} showFormControls={false} ... />}
+/>
+```
+
+|Prop|Type|Description|Default|
+|----|:--:|----------:|------:|
+|buttonClassName|string|Class which will be given to the buttons wrapper.|{ }|
```
Hyperkid123: This one is missing `onReset`, `onCancel` prop

packages/react-renderer-demo/src/app/pages/renderer/field-provider.md
```diff
@@ -9,47 +9,27 @@ import ListOfContents from '../../src/helpers/list-of-contents';
...
 Next example shows simple input field with label and error message.
 
 ```jsx
 import React from 'react';
+import { FieldProvider } from '@data-driven-forms/react-form-renderer'
 
-const NewComponent = ({ FieldProvider, formOptions, name, ...rest }) => (
+const NewComponent = ({ name, ...rest }) => (
   <div>
     <FieldProvider {...rest} name={name}>
```
Hyperkid123: FieldProvider no longer supports children. I would also rather show `useFieldApi`

packages/react-renderer-demo/src/app/pages/renderer/validators.md
```diff
@@ -56,6 +58,57 @@ Validator inputs and results are being cached so you will get immediate feedback
 If you do not want to trigger the async validator after every stroke, you can use a debounce promise [library](https://github.com/slorber/awesome-debounce-promise)
 (or any other implementation of debouncing.)
 
+# validatorMapper
```
Hyperkid123: Use nicer title please

packages/react-renderer-demo/src/app/pages/renderer/migration-guide.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Removed 'treshold' option
+
+-   Due to a typo, it was possible to use key treshold as an option in some validators, this option is now disabled. Use only threshold.
+
+### Form Template
+
+-   layoutMapper removed - instead of it use FormTemplate
+-   All props related to form buttons are now removed, use them in you FormTemplate
+    -   showFormControls
```
Hyperkid123: missing props for buttons

packages/react-renderer-demo/src/app/pages/renderer/migration-guide.md
```diff
@@ -0,0 +1,152 @@
...
+
+```jsx
+    import { FormSpy } from  '@data-driven-forms/react-form-renderer';
+
+{onStateUpdate &&  <FormSpy  onChange={onStateUpdate} />}
+```
+
+### FieldProvider changed
+
+-   FieldProvider is no longer wrapping basic components and you cannot access him through props, instead of it use hook:
```
Hyperkid123: ents and you cannot access **it** (instead of him) through props

packages/react-renderer-demo/src/app/pages/renderer/get-started.md
rvsia: lol :D

packages/react-renderer-demo/src/app/pages/renderer/component-api.md
rvsia: It's actually there `Each mapper provided in `react-forms` provides a default API for standard components.` I will expand it

packages/react-renderer-demo/src/app/pages/renderer/form-template.md
rvsia: These are passed via formApi. They are still mentioned in the form renderer api.

packages/react-renderer-demo/src/app/pages/renderer/migration-guide.md
rvsia: :question:

packages/react-renderer-demo/src/app/pages/renderer/form-template.md
Hyperkid123: `onReset` and `reset` are different. You can extend the default `reset` function with `onReset`. Show some notification for instance.

packages/react-renderer-demo/src/app/pages/renderer/migration-guide.md
Hyperkid123: onCancel, labels, etc are now in formtemplate not renderer props

packages/react-renderer-demo/src/app/pages/renderer/form-template.md
rvsia: onReset and onCancel are passed to FormRenderer and the renderer is adding them to the context. And you access the context via useFormApi

packages/react-renderer-demo/src/app/pages/renderer/migration-guide.md
rvsia: onCancel is still in the renreder. 

buttonsLabels, etc. are mentioned in this list.

packages/react-renderer-demo/src/app/pages/renderer/form-template.md
Hyperkid123: I know. I am just saying that I did not really notice it anywhere when going through the docs. Maybe this is a good candidate for example on a standalone page?

rvsia: onCancel and onReset all still mentioned in the renderer-api props list.

I don't think we need a standalone page for that, I can add it to custom-buttons example

Hyperkid123: I know they are mentioned but it seems to be that it is buried under a lot of other props etc. An example of such a basic scenario is a must-have a feel.

packages/react-renderer-demo/src/app/src/doc-components/custom-buttons.js
```diff
@@ -36,22 +36,32 @@ const FormTemplate = ({ formFields, schema }) => {
             <Button disabled={submitting || !valid} style={{ marginRight: 8 }} type="submit" color="primary" variant="contained">
               Submit
             </Button>
-            <Button disabled={pristine} style={{ marginRight: 8 }} onClick={reset} variant="contained">
+            <Button disabled={pristine} style={{ marginRight: 8 }} onClick={() => { onReset(); reset(); }} variant="contained">
```
Hyperkid123: we have `canReset` prop to show/hide the reset button. Cancel should be also wrapper in condition

packages/react-renderer-demo/src/app/pages/renderer/form-template.md
rvsia: I added using of onCancel, onReset to custom-buttons.

However, I found that we should do two fixes:

However, I found that we should  enhance these function in the formRenderer context:

```jsx
onReset: () => { onReset && onReset(); reset(); }
onCancel: () => onCancel(values)
```

packages/react-renderer-demo/src/app/src/doc-components/custom-buttons.js
rvsia: I would rather keep this as a minimalist example. Our formTemplate is more complex, there are more props (buttonOrders, etc.), so the example could become too big (and I don't think that we should show users how to use conditions in JSX :D )... and anyone can look on the code of our templates.

packages/react-renderer-demo/src/app/pages/renderer/form-template.md
Hyperkid123: Feel free to do them here.

rvsia: I would rather to do it in separate PR, this is out of the scope. :D

PR #377, opened 2020-03-12
fix(renderer): convert initialValue dataType

### LINKED ISSUE #376: renderer: Initial value data type is not applied on the field.
**Scope**: Renderer

**Description**
The initial value data type is not applied in the field.

**Schema**
```jsx
{
  component: 'text-field',
  name: 'integer-type',
  dataType: 'integer',
  initialValue: '5'
}
```

The initial value should be converted before its set as an initial value of a field.

Hyperkid123:
closes #376

- [x] tests

### DIFF
--- a/packages/react-form-renderer/src/tests/form-renderer/form-renderer.test.js
+++ b/packages/react-form-renderer/src/tests/form-renderer/form-renderer.test.js
@@ -31,7 +31,7 @@ describe('<FormRenderer />', () => {
     layoutMapper = {
       [layoutComponents.BUTTON_GROUP]: ({ children }) => <div>{ children }</div>,
       [layoutComponents.BUTTON]: ({ label, bsStyle, ...rest }) => <button { ...rest }>{ label }</button>,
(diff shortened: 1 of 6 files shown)


packages/react-form-renderer/src/tests/form-renderer/form-renderer.test.js
```diff
@@ -150,5 +150,67 @@ describe('<FormRenderer />', () => {
...
+        { ...initialProps }
+        schema={ schema }
+        onSubmit={ values => onSubmit(values) }
+        renderFormButtons={ FormControls }
+      />);
+      wrapper.find('form').simulate('submit');
+      expect(onSubmit).toHaveBeenCalledWith({ 'initial-convert': [ 5, 3, 11, 999 ]});
+    });
+
+    it.only('should convert individual values in array of objects as initial value', () => {
```
rvsia: Only

Hyperkid123: I swear I deleted it

PR #380, opened 2020-03-17
breaking(renderer): Allow combining async validators with normal ones.

Hyperkid123:


packages/react-form-renderer/demo/form-fields-mapper.js
```diff
@@ -7,17 +7,19 @@ import useFieldApi from '../src/hooks/use-field-api';
...
-        <label>{label} &nbsp;</label>
-        <input {...input} {...rest} />
-        {meta.error && (
-          <div>
-            <span>{meta.error}</span>
-          </div>
-        )}
-      </div>
-    )}
+    render={({ input, meta, isVisible, label, helperText, isRequired, dataType, isDisabled, isReadOnly, ...rest }) => {
```
rvsia: return block change for no reason ?

Hyperkid123: its just a demo dude :smile:

rvsia: dude, I spent the last minute searching the difference :D

packages/react-renderer-demo/src/app/pages/renderer/validators.md
```diff
@@ -83,7 +83,7 @@ const schema = {
 
 ```
 
-It is designed to return functions returning functions, so you can easily cached or debounce results.
+Validator in a mapper must be a function which returns a functions. This makes validator easilly configurable (different messages for same validator).
```
rvsia: returns a function~~s~~

rvsia: easilly > easily

PR #384, opened 2020-03-17
fix(pf4): wizard predicts steps by default

rvsia:
Removes predictSteps from PF4 wizard. It's always true. (There is no reason to not show all steps in the navigation, when it's possible.)


Hyperkid123:
@rvsia looks good. Can you also update the migration guide?

rvsia:
@Hyperkid123 done!

PR #388, opened 2020-03-18
[V2] Stream line build output files structure.

Hyperkid123:


Hyperkid123:
@rvsia I think it's ready. Checking if the docs still works would be a priority probably.

Hyperkid123:
Yeah that is the one thing that I cannot figure out. I wonder if that is only in SSR mode.

rvsia:
@Hyperkid123  so should we merge this?

Hyperkid123:
I don't know. We will have to deal with it later. And we should probably switch to the pf3 version of date time picker.

rvsia:
@Hyperkid123 yeah, let’s merged it, it is still just a beta

PR #391, opened 2020-03-19
feat(renderer): add proper not, and, or conditions

### LINKED ISSUE #390: Support for more complex conditions
So far we can only do AND conditions and OR conditions with a static right-hand-side, e.g. `a == x || b == x`. However, this is not sufficient to produce more advanced conditions. There should be a support for an independent OR operator and probably a NOT operator as well. Having fully featured AND, OR and NOT operators would allow us to do basically anything.

rvsia:
Closes #390 

TODO: 
- [x] docs
- [x] schema validator update

Is it alright @Hyperkid123 @skateman ? If so, I will finish it after the next meeting

Adds
- or
- and
- not
- can be nested

(All old conditions still work!)

Examples (look on the tests)

```js
      condition = {
        and: [
          { and: [{ when: 'x', pattern: /true/ }, { when: 'z', is: 'true' }]},
          { or: [{ when: 'y', pattern: /true/ }, { when: 'a', is: 'true' }]},
        ],
      };

      condition = {
        or: [
          { not: { when: 'x', pattern: /true/ }},
          { not: { when: 'y', pattern: /true/ }},
          { not: { when: 'z', pattern: /true/ }},
        ],
      };
```

### DIFF
--- a/packages/react-form-renderer/src/form-renderer/condition.js
+++ b/packages/react-form-renderer/src/form-renderer/condition.js
@@ -6,52 +6,59 @@ import get from 'lodash/get';
 
 const isEmptyValue = (value) => typeof value === 'number' || value === true ? false : lodashIsEmpty(value);
 
-const Condition = ({ condition, children }) => {
-  const fieldCondition = (value, { is, isNotEmpty, isEmpty, pattern, notMatch, flags }) => {
-    if (isNotEmpty){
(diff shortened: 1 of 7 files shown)


skateman:
I love it, starts looking like a really badly designed programming language :trollface: but for real, it's probably the best solution that maintains backwards compatibility.

PR #397, opened 2020-03-23
[V2] Share functionality in common select

rvsia:
- loadOptions functionality moved to common select
- PF4 select is using common now
- PF3 can load options by searching

### DIFF
--- a/packages/common/src/select/index.js
+++ b/packages/common/src/select/index.js
@@ -20,10 +20,9 @@ const handleSelectChange = (option, simpleValue, isMulti, onChange) => {
     : onChange(sanitizedOption);
 };
 
-const selectProvider = (type) =>
-  ({
-    createable: CreatableSelect
-  }[type] || ReactSelect);
+const selectProvider = {
+  createable: CreatableSelect
+};
 
 const Select = ({
   invalid,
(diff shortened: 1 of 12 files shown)


packages/common/src/select/index.js
```diff
@@ -15,29 +20,143 @@ const handleSelectChange = (option, simpleValue, isMulti, onChange) => {
     : onChange(sanitizedOption);
 };
 
-class Select extends Component {
-  render() {
-    const { input, invalid, classNamePrefix, simpleValue, isMulti, pluckSingleValue, options, ...props } = this.props;
-    const { value, onChange, ...inputProps } = input;
+const selectProvider = (type) =>
```
Hyperkid123: I think we should make this just a simple object.

packages/common/src/select/index.js
```diff
@@ -15,29 +20,143 @@ const handleSelectChange = (option, simpleValue, isMulti, onChange) => {
...
+      dispatch({ type: 'updateOptions', payload: data });
+    });
+  };
+
+  useEffect(() => {
+    if (loadOptions) {
+      updateOptions();
+    }
+
+    dispatch({ type: 'mounted' });
```
Hyperkid123: This is not reliable, you should use `useRef` for cases like this. There is an example: https://gist.github.com/jaydenseric/a67cfb1b809b1b789daa17dfe6f83daa

Hyperkid123:
the PF3 shows no data text instead of loading text.

pf3-loading

rvsia:
@Hyperkid123 Everything is fixed, MUI is using the common select too!

Hyperkid123:
@rvsia something hapenned

Issue #399, opened 2020-03-25
[V2] technical debt


Hyperkid123:
- [x] form template expected order SET?
- [x] from template form spy props
- [x] remove vendor assets from mui mapper
- [x] missing exports in component mapper indexes
- [x] mui missing wizard buttons labels
- [x] mui form template inline styles
- [x] missing HTML tag variants in mui plain text (see pf3 plain text)
- [x] mui radio using scss instead of JSS
- [x] mui select text field props color?
- [x] mui sub form inline styles and providers
- [x] scrips for demo comment and update demo are not being used, delete them 
- [x] pf3 selectcub components are not using class name prefix
- [x] pf3 select duplicate css imports
- [x] pf3 wizard step buttons inline styling
- [x] pf3 wizard step buttons form options default prop
- [x] pf3 plain text template strings
- [x] pf3 radio onChange extra block
- [x] pf3 radio unused rest spread props
- [x] pf3 sub form providers
- [x] pf3 switch onChange send event?
- [x] pf3 tab error has inline style
- [ ] apply tab error on all mappers?
- [x] `textarea` vs `text-area` in mappers file name 
- [x] pf4 wizard step to function
- [ ] dual list select should be in all mappers (feat)
  - [ ] mobile redesign
- [x] prune form renderer props
- [x] renderer missing form template template string
- [x] renderer unused state
- [x] use component spy is not used
- [x] missing `-validator` suffix in number value validators
- [x] check pf4 wizard and debug combination

### Demo, non blocking
- [x] next/prev buttons position
- [x] delete wrapped in form group  
- [ ] split component API and generate from props
- [x] conditions examples
- [x] action mapper example
- [x] Field Array provider title is not really explicit
- [x] move migration guide one level up
- [x] stop with firebase hosting release :smile: 
- [x] fix karel analyzer

rvsia:
>  check pf4 wizard and debug combination

FormTemplate shloud be a reference

DON'T

```jsx
<FormRenderer
{...props}
FormTemplate={(props) => <FormTemplate {...props} showFormControls={false />}}
/>
```

DO

```jsx
// outside of the render
FormTemplateWrapper = (props) => <FormTemplate {...props} showFormControls={false />};
...
<FormRenderer
{...props}
FormTemplate={FormTemplateWrapper}
/>
```

Hyperkid123:
@rvsia we should be able to overcome anonymous components context. This is a mistake on our part.

PR #404, opened 2020-03-27
Fix demo technical debts

rvsia:
Part of #399

- [x] waits for https://github.com/karelhala/commit-analyzer-wildcard/pull/17

### Demo, non blocking
- [x] next/prev buttons position
- [x] delete wrapped in form group  
- [x] conditions examples
- [x] action mapper example
- [x] fix karel analyzer

Only thing I added as extra is a text to the raw component, I think it's better to tell users they can click and expand the code:

image

### DIFF
--- a/packages/react-renderer-demo/src/app/src/components/common/raw-component.js
+++ b/packages/react-renderer-demo/src/app/src/components/common/raw-component.js
@@ -20,13 +20,18 @@ const req = require.context('@docs/doc-components', true, /\.js/);
 const useStyles = makeStyles((theme) => ({
   heading: {
     fontSize: theme.typography.pxToRem(15),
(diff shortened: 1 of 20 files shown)


packages/react-renderer-demo/src/app/src/components/common/raw-component.js
```diff
@@ -20,13 +20,20 @@ const req = require.context('@docs/doc-components', true, /\.js/);
 const useStyles = makeStyles((theme) => ({
   heading: {
     fontSize: theme.typography.pxToRem(15),
-    fontWeight: theme.typography.fontWeightRegular
+    fontWeight: theme.typography.fontWeightRegular,
+    flexBasis: '33.33%',
```
Hyperkid123: what about 100 / 3 :wink:

PR #422, opened 2020-04-08
Add notification panel to demo

rvsia:
This PR adds a notification panel to demo
- last checked is stored in localStorage
- using MDX to format messages (can use links, etc.)
- messages are hard coded in the code

demonotifications


Hyperkid123:
@rvsia nice. I think we can think of some way to make the dynamic later.

I think it would be also worthwhile to add the latest version badge to the landing page.

rvsia:
@Hyperkid123 done

image

PR #432, opened 2020-04-16
feat(renderer): enhance conditions

### LINKED ISSUE #427: Allow Conditions to set form values
**Scope**:  FormRender, Conditions

**Description**

User may often want to use a condition to not only (optionally) evaluate the display criteria for a form field, but to also (optionally) set a form value. For example, if there is a simple 3 field form - field A is a number, field B is a string and field C should have its value set based on a set of conditions on both A and B.

This feature should accomodate the ability to

- evaluate a condition and set a form field without changing the visibility of the  field
- evaluate a condition and set a form field and change the visibility of the field
- support hidden, but enabled fields

**Semantics / Syntax Suggestion**

**<u>Evaluate a condition and set a form field without changing the visibility of the field</u>**

```javascript
{
  fields: [
    {
      name: 'Foo', // controlled field
      component: 'text-field',
      condition: {   // <-- new keyword 
        when: 'BarFoo', // name of controlled field
        is: 'Bar',      // condition
        remainVisible: true // remain visible (default)
        then: [{ // <-- absense of the "set" keyword means to set this field
          to: 'I have been set because BarFoo = Bar' // value to set
        },
        {
	 set: 'Woo',                                // name of field to set
          to: 'I have been set because BarFoo = Bar' // value to set
        }]
      }
    }, 
    {
      name: 'Woo', // controlled field
      component: 'text-field',
    },
    {
      name: 'BarFoo',
      label: 'Foo is Bar!',
      component: 'text-field',
    }
  ]
}
```

**<u>Evaluate a condition and set a form field and change the visibility of the field</u>**

```javascript
{
  fields: [
    {
      name: 'Foo', // controlled field
      component: 'text-field',
      condition: {   // <-- new keyword 
        when: 'BarFoo', // name of controlled field
        is: 'Bar',      // condition
        remainVisible: false // become visible when condition is met and set fields
        then: [{ // <-- absense of the "set" keyword means to set this field
          to: 'I have been set because BarFoo = Bar' // value to set
        }]
      }
    }, 
    {
      name: 'Woo', // controlled field
      component: 'text-field',
    },
    {
      name: 'BarFoo',
      label: 'Foo is Bar!',
      component: 'text-field',
    }
  ]
}
```

rvsia:
- adds sequence condition
- adds then, else branches
- adds set action
- optimize condition to listen values changes

close #427 

TODO:
- [x] tests
- [x] schema validator update
- [x] documentation
- [x] first render issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     }
   },
   "devDependencies": {
+    "@babel/plugin-transform-runtime": "^7.9.0",
     "@khala/commit-analyzer-wildcard": "^2.4.1",
     "@khala/npm-release-monorepo": "^2.4.1",
     "@khala/wildcard-release-notes": "^2.4.1",
--- a/packages/react-form-renderer/src/form-renderer/condition.js
+++ b/packages/react-form-renderer/src/form-renderer/condition.js
@@ -1,8 +1,10 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import lodashIsEmpty from 'lodash/isEmpty';
-import { FormSpy } from 'react-final-form';
 import get from 'lodash/get';
+import isEqual from 'lodash/isEqual';
+
+import useFormApi from '../files/use-form-api';
 
 const isEmptyValue = (value) => (typeof value === 'number' || value === true ? false : lodashIsEmpty(value));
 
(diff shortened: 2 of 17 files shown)


packages/react-form-renderer/src/form-renderer/condition.js
```diff
@@ -27,48 +29,130 @@ const fieldCondition = (value, { is, isNotEmpty, isEmpty, pattern, notMatch, fla
...
+            });
+          });
+        }
+      });
+      setSets(setters);
+    }
+  }, [setters, dirty]);
+
+  return conditionResult.visible ? children : null;
+}, isEqual);
```
Hyperkid123, line 121: I think this memo comparator is not going to work very well. Using just the `values` slightly more efficient. This way you compare all the children nodes of condition.

Hyperkid123: It should mainly prevent re-rendering the FormSpy wrapper which is the most expensive component.

packages/react-form-renderer/src/form-renderer/condition.js
```diff
@@ -27,48 +29,137 @@ const fieldCondition = (value, { is, isNotEmpty, isEmpty, pattern, notMatch, fla
...
+            setTimeout(() => {
+              formOptions.batch(() => {
+                Object.entries(setter).forEach(([name, value]) => {
+                  formOptions.change(name, value);
+                });
+              });
+            });
+          }
+        });
+        setSets(setters);
```
Hyperkid123, line 120: we are doing two setters in one effect, that will cause additional render. Can we move from `useState` to `useReducer`?

package.json
```diff
@@ -37,6 +37,7 @@
     }
   },
   "devDependencies": {
+    "@babel/plugin-transform-runtime": "^7.9.0",
```
Hyperkid123, line 40: is this new plugin necessary? I think that `babel-env` should provide same features.

rvsia:
@Hyperkid123 

`useState` transformed to `useReducer`

babel-env is getting targets to enable async/await generators

Hyperkid123:
> @Hyperkid123
> 
> `useState` transformed to `useReducer`
> 
> babel-env is getting targets to enable async/await generators

Riight that's it. :thinking: So it affects SSR and node dependency. Hm, I am thinking that we should rather use the plugin then. If might become an issue if somebody tries to deploy this on SSR env with an outdated version of node.

rvsia:
@Hyperkid123 like next.js on firebase, right? :D 

Undropped!

PR #436, opened 2020-04-17
fix(pf4): refactor wizard, validating disable nav

rvsia:
- refactors code, make it more simple and cleaner
- fix bugs when there are some fields on the last step
- disabling navigation when validating is in progress
- optimize subscription more

validating

### DIFF
--- a/packages/pf4-component-mapper/src/files/wizard.js
+++ b/packages/pf4-component-mapper/src/files/wizard.js
@@ -1,4 +1,4 @@
-import React, { cloneElement, useReducer, useEffect } from 'react';
+import React, { useReducer, useEffect } from 'react';
 import { createPortal } from 'react-dom';
 import PropTypes from 'prop-types';
 import { FormSpy } from '@data-driven-forms/react-form-renderer';
@@ -72,10 +72,6 @@ const WizardInternal = ({
     return null;
   }
 
-  const step = (
-    <WizardStep {...currentStep} formOptions={formOptions} buttonLabels={buttonLabels} buttonsClassName={buttonsClassName} showTitles={showTitles} />
-  );
-
   return (
     <Modal inModal={inModal} container={state.container}>
       <div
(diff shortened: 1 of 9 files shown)


packages/pf4-component-mapper/src/files/wizard.js
```diff
@@ -89,27 +85,33 @@ const WizardInternal = ({
         {title && <WizardHeader title={title} description={description} onClose={formOptions.onCancel} />}
         <div className="pf-c-wizard__outer-wrap">
           <WizardNav>
-            <FormSpy>
-              {({ values }) => (
+            <FormSpy subscription={{ values: true, valid: true, validating: true }}>
```
Hyperkid123, line 88: what about submitting? I think we should block the navigation while submitting as well.

packages/pf4-component-mapper/src/files/wizard/wizard-nav.js
```diff
@@ -20,19 +20,21 @@ const memoValues = (initialValue) => {
...
 
 const WizardNavigationInternal = React.memo(
-  ({ navSchema, activeStepIndex, maxStepIndex, jumpToStep, formOptions: { valid } }) =>
+  ({ navSchema, activeStepIndex, maxStepIndex, jumpToStep, valid, validating }) =>
     navSchema
       .filter((field) => field.primary)
       .map((step) => {
         const substeps = step.substepOf && navSchema.filter((field) => field.substepOf === step.substepOf);
 
+        const isValid = valid && !validating;
```
Hyperkid123, line 29: ^^

packages/pf4-component-mapper/src/files/wizard.js
rvsia: the whole navigation? this just disables next steps (on the last step there are no next steps)

Hyperkid123: Is it ok to move in the wizard when the data is in the submission process? I think the navigation in the wizard should be postponed.

rvsia: After submit, the wizard form should go away and be replace by spinner/loading progress according to design specs - there should be no navigation or buttons at all.

rvsia:
+ I am also fixed a little bug in the reducer, when users click on the current step on the initial page (there wasn't any history, so the wizard failed.)

packages/pf4-component-mapper/src/files/wizard.js
Hyperkid123: Yes but we are talking about the time period when the submit is in progress not when it was finished.

PR #442, opened 2020-04-21
fix(renderer): move functionality from render to useField

rvsia:
Some functionality left in render single field, but it should be moved to useField to make sense.

- `renderSingleField` does not do anything, only renders component and use action mapper to assign props
- `useField` now creates the validate, validatorArray, converting initialValue
- fixes some propTypes
- fixes wrong initializeOnMount tests
- **breaking**: FieldProvider uses `Component` instead of `component` (due to props naming conflict)

### DIFF
--- a/packages/pf3-component-mapper/src/tests/__snapshots__/input-addon.test.js.snap
+++ b/packages/pf3-component-mapper/src/tests/__snapshots__/input-addon.test.js.snap
@@ -369,6 +369,7 @@ exports[`<Input Addon> should render group button before input addon correctly 1
                       hideField={false}
                     >
                       <TextField
+                        component="text-field"
                         inputAddon={
                           Object {
                             "before": Object {
@@ -523,6 +524,7 @@ exports[`<Input Addon> should render group button before input addon correctly 1
                                           hideField={false}
                                         >
                                           <InputAddonButtonGroup
+                                            component="input-addon-button-group"
                                             fields={
                                               Array [
                                                 Object {
@@ -558,6 +560,7 @@ exports[`<Input Addon> should render group button before input addon correctly 1
                                                       hideField={false}
                                                     >
                                                       <Button
+                                                        component="button"
                                                         label="Set 1"
                                                         name="set1"
                                                         onClick={[Function]}
@@ -567,12 +570,14 @@ exports[`<Input Addon> should render group button before input addon correctly 1
                                                           block={false}
                                                           bsClass="btn"
                                                           bsStyle="default"
+                                                          component="button"
                                                           disabled={false}
                                                           name="set1"
                                                           onClick={[Function]}
                                                         >
                                                           <button
                                                             className="btn btn-default"
+                                                            component="button"
                                                             disabled={false}
                                                             name="set1"
                                                             onClick={[Function]}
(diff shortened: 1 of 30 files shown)


packages/react-form-renderer/src/files/use-field-api.js
```diff
@@ -1,38 +1,107 @@
...
+        setType(specialType);
+      }
+    }
+  }, [component]);
+
+  /** Reinitilize array validator/validate */
+  useEffect(() => {
+    if (mounted.current) {
+      setArrayValidator(calculateArrayValidator(props, validate, component, validatorMapper));
+      setValidate(calculateValidate(props, validate, component, validatorMapper));
```
Hyperkid123, line 61: we could also refactor to `useReducer` if we want reduce the number of renders.

packages/react-form-renderer/src/files/use-field-api.js
```diff
@@ -1,38 +1,107 @@
...
+    }
+  }, [validate, component, props.dataType]);
+
+  /** Re-convert initialValue when changed */
+  useEffect(() => {
+    if (mounted.current) {
+      const newInitialValue = calculateInitialValue(props);
```
Hyperkid123, line 68: we should probably count this value only if the condition for the state update is met. We can save some resources.

packages/react-form-renderer/src/files/use-field-api.js
```diff
@@ -1,38 +1,107 @@
...
-      }
+    () => {
+      mounted.current = true;
+
+      return () => {
+        /**
+         * Delete the value from form state when field is inmounted
+         */
+        if ((formOptions.clearOnUnmount || props.clearOnUnmount) && props.clearOnUnmount !== false) {
+          fieldProps.input.onChange(fieldClearedValue);
```
Hyperkid123, line 102: we should also set mounted to `false` before unmount

packages/pf3-component-mapper/src/tests/__snapshots__/input-addon.test.js.snap
```diff
@@ -567,12 +570,14 @@ exports[`<Input Addon> should render group button before input addon correctly 1
...
                                                           bsClass="btn"
                                                           bsStyle="default"
+                                                          component="button"
                                                           disabled={false}
                                                           name="set1"
                                                           onClick={[Function]}
                                                         >
                                                           <button
                                                             className="btn btn-default"
+                                                            component="button"
```
Hyperkid123, line 580: the component props seem to have sneak to actual DOM elements. We should fix that.

packages/react-form-renderer/src/tests/form-renderer/__snapshots__/render-form.test.js.snap
```diff
@@ -452,9 +460,11 @@ exports[`renderForm function should render single field from defined componentTy
           hideField={false}
         >
           <Component
+            component="text-field"
             name="foo"
           >
             <div
+              component="text-field"
```
Hyperkid123, line 467: again, we have the component on the DOM element.

rvsia: This is actually right. It's just stupid <div {...props} /> component :D 

I caught the prop in all other proper components.

packages/react-form-renderer/src/files/use-field-api.js
rvsia: Not sure if we can make it more conditional. Right now, it's being recalculated when initial value or datatype are changed. Datatype's change should definitely recalculate the value... initialValue only if the value is not of the it's type, so we could add some check function (and also added it to the first calculation)

PR #450, opened 2020-04-23
fix(pf4): allow to pass container to wizard

rvsia:
Allow to pass a container to the wizard, so we can hide it in DOM if we need to


Hyperkid123:
released in: @data-driven-forms/pf4-component-mapper@1.31.2

PR #452, opened 2020-04-24
fix(pf4): add styling for disabled options

rvsia:
selectdisable

Disabled option was missing styling, done according to PF select component.


Hyperkid123:
released in @data-driven-forms/pf4-component-mapper@1.31.3

PR #458, opened 2020-04-27
fix(all): enable input type file

### LINKED ISSUE #307: PF3: File upload component
Just tracking that this is necessary for the provider forms.

Hyperkid123:
closes #307

this will have to be backported to V1

### To Do
- [x] tests
- [x] docs
  - [x] validation example


skateman:
Can't really say anything useful without an example usage piece of code :thinking:

Hyperkid123:
@skateman I added examples to the docs. Source starts from here: https://github.com/data-driven-forms/react-forms/pull/458/files#diff-63eddbe3b4330a752e2fa0ffad9dba2fR1

packages/react-renderer-demo/src/app/pages/renderer/file-input.md
```diff
@@ -0,0 +1,40 @@
...
+
+import ListOfContents from '../../src/helpers/list-of-contents';
+import CodeExample from '../../src/components/code-example';
+
+<Grid container item>
+<Grid item xs={12} md={10}>
+
+# File input
+
+Files cannot be easilly uploaded in JSON payload. In order to upload files usigng the input type *file* you can follow these steps.
```
rvsia, line 12: easilly > easily

usigng > using

packages/react-renderer-demo/src/app/pages/renderer/file-input.md
```diff
@@ -0,0 +1,40 @@
...
+
+## File onChange payload
+
+In order to successfully store the file reference, you have either use an input of type file or use an object with the following shape in your on change function ([visit MDN docs for more info on file upload](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file)).
```
rvsia, line 16: maybe: 'use an input of "file" type'

packages/react-renderer-demo/src/app/pages/renderer/file-input.md
```diff
@@ -0,0 +1,40 @@
...
+  target: {
+    value: string, // usually it is the path to the file
+    files: FileList // an array of file references
+  }
+}
+```
+
+## Getting file from state.
+
+When submitting, you will have to construct the binary via FormData or encode the file to Base64, depending on your use case. Be aware that FormData cannot be sent in the JSON payload. Binaries are destroyed when serializing JSON. There will be a list of field names with the type file avaiable in the submit function arguments.
```
rvsia, line 28: avaiable > available

packages/react-renderer-demo/src/app/pages/renderer/file-input.md
```diff
@@ -0,0 +1,40 @@
...
+
+The `formApi.fileInputs` is an array of field names with `type: file`. Be aware that if your filed name is in nested, you have to use the lodash like method of getting the value from state.
```
rvsia, line 30: filed > field

packages/react-form-renderer/src/files/use-field-api.js
```diff
@@ -76,6 +76,9 @@ const useFieldApi = ({ name, initializeOnMount, component, render, validate, ...
   };
 
   const fieldProps = useField(name, enhancedProps);
+  if (fieldProps.input.type === 'file' && typeof fieldProps.input.value === 'object') {
```
rvsia, line 79: I think this modification should be moved to the return statement of this hook.

Hyperkid123: why?

rvsia: With this the input object is basically being modified at two different places. I would keep it at one place - it would be more consistent and readable.

Hyperkid123: its not going to be pretty. But ok

packages/react-form-renderer/src/files/use-field-api.js
```diff
@@ -170,6 +177,8 @@ const useFieldApi = ({ name, initializeOnMount, component, render, validate, ...
     ...(arrayValidator ? { arrayValidator } : {}),
     input: {
       ...fieldProps.input,
+      value:
```
rvsia, line 180: What about?
`...(fieldProps.input.type === 'file' && typeof fieldProps.input.value === 'object' && {value: fieldProps.input.value.inputValue})`

Hyperkid123: that's how I originally wrote it but linter did it this way

rvsia: :man_facepalming:

davellx:
I was trying to make my own component for file inputs but couldn't go past the enhancedOnChange at the moment so I'm glad this PR exists, thank you ! 
Maybe It should be a good thing to plug some of the basic validation (required, length ?) or to explain in the docs that validation mut be done manually for future users.

Hyperkid123:
> I was trying to make my own component for file inputs but couldn't go past the enhancedOnChange at the moment so I'm glad this PR exists, thank you !
> Maybe It should be a good thing to plug some of the basic validation (required, length ?) or to explain in the docs that validation mut be done manually for future users.

The validation of input files is a bit tricky compared to other components. I think adding an example with a custom validator is a good idea. :+1:

Hyperkid123:
@davellx @rvsia I have added validation to the docs example. We can think about proper file validator in different PR. There are enough tools available to make the file upload work as it is and I don't want to block it too much.

PR #460, opened 2020-04-28
Fix styling issues on documentation

rvsia:
**Width on smaller screen**

CSS has been broken for so long times on smaller screen

Before

phonebefore

After

phoneafter

**Right footer is being shrink**

Before

image

After

image

**Make footer different color**

Before

image

After

image

**Make navlink buttons be more consistent part of the layout**

Before

image

After

image


Hyperkid123:
@rvsia while you are at this. Could you make the nav to take the whole screen width when on small devices?

PR #463, opened 2020-04-30
Migrate docs to MUI and update component examples.

Hyperkid123:
### Changes
- remove PF from docs build
  - whole docs is now using MUI
- offload component examples to stackblitz embedded editor
  - allows us to reduce the docs build size and loading times (from 15MB to 3.4 MB) because we no longer need to build un optimized libraries
  - allow further scaling of component mapper examples
  - users can also edit examples instead of just changing props
- great performance improvements
  - build size is now just under 3.5MB for the client
  - server has now around 10MB
  - biggest initial load is 260KB
  - removed CSS compilation from the project, any CSS files can be included only in the `_app.js` file 

### Component example changes
swapmapper gif


Hyperkid123:
Current progress: 
swapmapper

Now just have to protect the layout from global styling.

Hyperkid123:
Upon further consideration, I don't believe that scoping the docs for specific component library is a good idea. As the number of supported libraries going to grow, the build size of the docs will grow very quickly, especially if the library is not properly optimized for SSR. Adding PF4 increased the server size from 10MB to 40MB.

So I think we should have the docs just in MUI and then have embedded codesanbox for other libraries so we can offload the build to their side. @rvsia I believe you will be happy about this.

rvsia:
I am sorry to hear about it... but this will be probably much easier to do and more future-proof.

Hyperkid123:
@rvsia I think most of the features are done. Now its time for cleaning the docs up. Can you take a look? I have lost track already.

rvsia:
Maybe we could switch to light theme if it's possible

image

It fits the page better I think. :D

rvsia:
Is it possible to add some loading indicator?

loading

Hyperkid123:
@rvsia Added responsiveness and a cheeky spinner to indicate some loading.

I will also disable the navigation opening on each link click.

responsive

Hyperkid123:
@rvsia I dont like the light theme :disappointed:

rvsia:
I think you missed the comment about missing texts for other mappers. :(

example:
1. Go to Wizard
2. Select PF4
3. There is no documentation text

Hyperkid123:
@rvsia all right the text is back.

PR #471, opened 2020-05-04
Add MUI dual-list-select component

rvsia:
TODO:
- [x] PF4 dualllist > common
- [x] MUI dual list
- [x] tests
- [x] update demo
- [x] renderer make dual list select constant

example (checkbox variant)

muiduallist

### DIFF
--- a/packages/common/src/dual-list-select/index.js
+++ b/packages/common/src/dual-list-select/index.js
@@ -1,0 +1,102 @@
+import React, { useReducer } from 'react';
+import PropTypes from 'prop-types';
+import isEqual from 'lodash/isEqual';
+
+import { useFieldApi } from '@data-driven-forms/react-form-renderer';
+
+import reducer, { initialState } from './reducer';
+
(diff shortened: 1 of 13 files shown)


Hyperkid123:
@rvsia I think we should probably use adornments for the icons: https://material-ui.com/components/text-fields/#input-adornments

rvsia:
@Hyperkid123 The input is filtering, the icon is sorting.  Not sure if we should mix these two actions in one input.

Hyperkid123:
:thinking: hmm that icon does not remind me of sorting

rvsia:
@Hyperkid123 It's literally called 'SortIcon' :smile: I can use arrow if you want to.

Hyperkid123:
I know but it just does not seem right to me :smiley:. If you look at the table demo for example, that are using the arrows as well: https://material-ui.com/components/tables/#sorting-amp-selecting

rvsia:
arrow

done

PR #474, opened 2020-05-06
Make menu on small screen temporary

rvsia:
On small screen the menu is temporary and the content is not shifted

mobilephone

On medium+ screens the menu is persistent (business as usual) 

standard


Hyperkid123:
@rvsia i thingk we should fix up the spacing around the search input. It overlaps the menu button on most of the mobile screens (I've tried several presets in chrome):
screenshot-localhost_3000-2020 05 06-11_58_16

rvsia:
@Hyperkid123 added paddingLeft to the bar

image

PR #489, opened 2020-05-14
fix(common): make props to control deselection of values

rvsia:
fix for the bug found in https://github.com/RedHatInsights/approval-ui/pull/233

This PR adds opt-in prop to disable `value` manipulation when the selected value is not in new options (after the props is changed, not when filtering!)

It would make more sense to make it opt-out, however, this feature was made to allow editing options in editors and it could introduce breaking changes for someone. What do you think, @Hyperkid123 ?

### DIFF
--- a/packages/common/src/select/index.js
+++ b/packages/common/src/select/index.js
@@ -40,6 +40,7 @@ const Select = ({
   value,
   onChange,
   loadOptionsChangeCounter,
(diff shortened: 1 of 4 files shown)


Hyperkid123:
We are using similar work around in the catalog aswell i was not usre if this should be included in here but it looks like it should

PR #490, opened 2020-05-15
Suir customization

### LINKED ISSUE #484: [v2.4] SUIR customization of inner components
The first iteration of mapper does not allow any composite components customization. Fix that.

Hyperkid123:
closes #484

### DIFF
--- a/packages/suir-component-mapper/src/common/form-field-grid.js
+++ b/packages/suir-component-mapper/src/common/form-field-grid.js
@@ -11,12 +11,12 @@ const useStyles = createUseStyles({
   }
 });
 
-const FormFieldGrid = ({ className, children, helperText }) => {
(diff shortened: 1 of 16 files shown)


packages/suir-component-mapper/src/common/form-field-grid.js
```diff
@@ -11,20 +11,21 @@ const useStyles = createUseStyles({
   }
 });
 
-const FormFieldGrid = ({ className, children, helperText }) => {
+const FormFieldGrid = ({ className, children, helperText, helperTextClassName }) => {
```
rvsia, line 14: what about do `HelperTextProps` instead of only className?

Hyperkid123: Right

PR #500, opened 2020-05-20
Update packages readme files.

Hyperkid123:


packages/common/README.md
```diff
@@ -4,30 +4,24 @@
 ￼
 This package exists to share code used by more than one data-driven-forms package, such as:
 
-* shared interfaces - propTypes
-* common mapper code - `condition` implementation
+- shared interfaces - propTypes
+- common mappers code 
+  - wizard
+  - select
+  - field array
```
rvsia, line 11: actually there is no common field-array yet

packages/react-form-renderer/README.md
```diff
@@ -26,15 +26,11 @@ Data Driven Forms is a React library used for rendering and managing forms with
 
 :book: For more information please visit the [documentation](https://data-driven-forms.org/). :book:
 
-Used by [ManageIQ](http://manageiq.org/), Red Hat Cloud Services.
 
 **Table of Contents**
 
 - [Installation](#installation)
   - [React Form Renderer](#react-form-renderer)
-  - [PatternFly 3 Mapper](#patternfly-3-mapper)
```
rvsia, line 35: What about to add 'mappers' section here?

packages/common/README.md
Hyperkid123: You are right.

packages/react-form-renderer/README.md
Hyperkid123: I don't think we need it here. Its mentioned in the root Readme and each mapper has its own section. This should be purely used for the renderer.

rvsia: The readme is mirrored into NPM, so when somebody comes from it, it would be nice to show him that there are several mapper he can use.

Hyperkid123: Shoot you are right. It's this one.

PR #503, opened 2020-05-21
fix(mui): use mui select instead of react select

### LINKED ISSUE #478: [v2.4] Remove react-select from MUI
With the changes to the common select and new select components in MUI, we can start using all the DDF features without the react-select library if the mapper library has good select component

Hyperkid123:
closes #478


packages/mui-component-mapper/demo/index.js
```diff
@@ -29,8 +30,118 @@ const compositeMapper = {
...
+      component: componentTypes.SELECT,
+      label: 'Async select',
+      loadOptions
+    }
+  ]
+};
+
 const App = () => {
-  const [schema, setSchema] = useState(wizardSchema);
+  const [, setSchema] = useState(wizardSchema);
```
rvsia, line 144: I would keep this, just add a new button with the schema

packages/mui-component-mapper/src/files/select.js
```diff
@@ -3,39 +3,108 @@ import PropTypes from 'prop-types';
...
+              readOnly: !isSearchable,
+              placeholder: !internalValue || internalValue.length === 0 ? placeholder : undefined
+            }}
+          />
+        )}
+        noOptionsText={noOptionsMessage()}
+        onInputChange={(_event, value) => onInputChange(value)}
+        options={options}
+        multiple={isMulti}
+        getOptionLabel={(option) => {
```
rvsia, line 73: can you extract this function?

packages/mui-component-mapper/src/files/select.js
```diff
@@ -3,39 +3,108 @@ import PropTypes from 'prop-types';
...
+
+          const item = options.find(({ value }) => value === option);
+          return (item && item.label) || '';
         }}
-        input={input}
-        label={label}
-        helperText={invalid || helperText || description}
-        {...rest}
+        value={typeof internalValue === 'undefined' ? '' : internalValue}
+        onChange={(_event, option) => {
```
rvsia, line 94: ^^ it would allow better testability and readability

packages/mui-component-mapper/src/files/select.js
```diff
@@ -44,15 +113,66 @@ Select.propTypes = {
...
+  isFetching: PropTypes.bool,
+  noOptionsMessage: PropTypes.func,
+  closeMenuOnSelect: PropTypes.bool,
+  onChange: PropTypes.func.isRequired,
+  hideSelectedOptions: PropTypes.bool,
+  required: PropTypes.bool,
+  onFocus: PropTypes.func,
+  onBlur: PropTypes.func,
+  TextFieldProps: PropTypes.object,
+  inputProps: PropTypes.object
```
rvsia, line 132: What is the difference between input props and TextField props?

Hyperkid123: They are different components. TextField controls the labels colors, helper text, etc, but the input props control the actual input.

Its been done based in MUI docs.

Hyperkid123:
> I suppose there is a prop, that will close the menu

Yes there is and it should be closed by default. Probably wring negation somewhere in the props bridge.

Hyperkid123:
@rvsia fixed

PR #513, opened 2020-05-22
Release 2.4

Hyperkid123:
Before merging we have to manually update firebase functions because of the name change.

New packages should be released automatically because we already have beta builds published.

I will double-check the docs locally


Hyperkid123:
:pray: lets GO

PR #515, opened 2020-05-25
fix(all): include prop-types in dependencies

Hyperkid123:
`PropTypes.elementType` cause consumer apps to crash if they used an older version of `prop-types` package.


packages/suir-component-mapper/package.json
```diff
@@ -58,6 +57,7 @@
   },
   "peerDependencies": {
     "@data-driven-forms/react-form-renderer": "^2.2.0",
+    "prop-types": "^15.7.2",
```
rvsia, line 60: peerDependency

PR #525, opened 2020-05-26
[TS] convert react form renderer to TS

### LINKED ISSUE #519: [TS] Add TS to react form renderer package
- types
- props
This should enable better DX and some prop types generation in docs

Hyperkid123:
closes #519


Hyperkid123:
The basic migration is done
Every js file in renderer package has now `.d.ts` file.
Now they have to be checked and tested in prod.

### TO DO
- [ ] have `index.d.ts` in root folders
- [ ] have on `index.d.ts` in the env root folder
- [ ] check what kinds `.d.ts` files should be in the build bundle

Hyperkid123:
Ok I think that everything for initial TS support for renderer is done. I will create a beta release and do some testing inside TS project.

Hyperkid123:
After initial testing, there are these problems with types:
- [ ] no types for FormTemplate
- [ ] useFormApi does not extend final form API methods
- [ ] FormSpy is not exported as react component
- [ ] validate in schema points to the wrong typings
- [ ] component mapper: element is not assignable to type string
- [ ] useFieldApi is not extensible with additional attributes (input and meta is OK)
- [ ] onSubmit does only accept a callback that returns a promise

More work!

Hyperkid123:
@rvsia I think I am done. I believe all the typings are correct. I was testing it locally and going through the schema and component props to make sure the definitions are correct.

We will also have some time to test it while adding types to component mappers.

It should be ready for review.

PR #537, opened 2020-06-05
[TS] Mui typings

### LINKED ISSUE #524: [TS] MUI

Hyperkid123:
closes #524 

requires #538 

I have found a missing type in renderer for a use field API params outside of schema, the validate param should be a function instead of array of objects

I have to add that before merging this PR.


Hyperkid123:
- [ ] Test types in TS project

Hyperkid123:
@rvsia ok, the typings seems to be working well

PR #538, opened 2020-06-05
[TS] fix(renderer): export useFieldApiConfig for components.

Hyperkid123:


Hyperkid123:
Its get copied to index in build

PR #540, opened 2020-06-08
[TS] added types for pf4 mapper

### LINKED ISSUE #523: [TS] PF4

Hyperkid123:
closes #523

### DIFF
--- a/packages/pf4-component-mapper/src/files/field-array.d.ts
+++ b/packages/pf4-component-mapper/src/files/field-array.d.ts
@@ -1,0 +1,16 @@
+import { ReactNode } from "react";
+import { Field } from "@data-driven-forms/react-form-renderer";
+
+export interface FieldArrayProps {
+  label?: ReactNode;
+  description?: ReactNode;
+  fields: Field[];
+  defaultItem?: any;
+  minItems?: number;
+  maxItems?: number;
+  noItemsMessage?: ReactNode;
+}
+
+declare const FieldArray: React.ComponentType<FieldArrayProps>;
+
+export default FieldArray;
--- a/packages/pf4-component-mapper/src/files/field-array.d.ts
+++ b/packages/pf4-component-mapper/src/files/field-array.d.ts
@@ -1,10 +1,10 @@
 import { ReactNode } from "react";
-import { Field } from "@data-driven-forms/react-form-renderer";
+import { FieldArrayField } from "@data-driven-forms/react-form-renderer";
 
 export interface FieldArrayProps {
   label?: ReactNode;
   description?: ReactNode;
-  fields: Field[];
+  fields: FieldArrayField[];
   defaultItem?: any;
   minItems?: number;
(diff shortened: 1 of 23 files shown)


packages/pf4-component-mapper/src/files/wizard.d.ts
```diff
@@ -0,0 +1,36 @@
...
+  description?: ReactNode;
+  isCompactNav?: boolean;
+  inModal?: boolean;
+  setFullWidth?: boolean;
+  setFullHeight?: boolean;
+  isDynamic?: boolean;
+  showTitles?: boolean;
+  crossroads?: string[];
+  fields: Field[];
+  navSchema?: NavSchemaStep[];
```
rvsia, line 31: navSchema is passed by internal common wizard

packages/pf4-component-mapper/src/files/wizard.d.ts
```diff
@@ -0,0 +1,28 @@
...
+  title?: ReactNode;
+  description?: ReactNode;
+  isCompactNav?: boolean;
+  inModal?: boolean;
+  setFullWidth?: boolean;
+  setFullHeight?: boolean;
+  isDynamic?: boolean;
+  showTitles?: boolean;
+  crossroads?: string[];
+  fields: Field[];
```
rvsia, line 23: It isn't a field, it shoul be something like WizardStep

```ts
WizardStep  {
   name: String;
   nextStep: String | Object | Function;
   substepOf: String;
   fields: Field[];
   ... (please see https://data-driven-forms.org/component-example/wizard?mapper=pf4 # Docs for steps )
}
```

packages/pf4-component-mapper/src/files/field-array.d.ts
```diff
@@ -0,0 +1,16 @@
+import { ReactNode } from "react";
+import { Field } from "@data-driven-forms/react-form-renderer";
+
+export interface FieldArrayProps {
+  label?: ReactNode;
+  description?: ReactNode;
+  fields: Field[];
```
rvsia, line 7: Oh.. does field requires `name`? FieldArray field if there is only one, does not require the name.

packages/pf4-component-mapper/src/files/wizard.d.ts
Hyperkid123: yeah name is required always in `Field` type

packages/pf4-component-mapper/src/files/field-array.d.ts
Hyperkid123: yeah name is required always in `Field` type

rvsia: So then it should not be field :D

Hyperkid123: @rvsia I've added special type for the field array field

PR #541, opened 2020-06-08
Add selected items status message to dual list

boaz0:
I want to add the ability to pass a selected items status to the dual list.

This is a draft:
- [x] PF4 - updated the docs too
- [x] PF3 - I will pass.
- [x] BJS - I will pass as I realized that it is not a must and each mapper has its own props
- [x] MUI - I will pass as I realized that it is not a must and each mapper has its own props
- [x] SUIR - I will pass as I realized that it is not a must and each mapper has its own props

The message will have placeholders `[selected]` and `[options]` to put the number of selected items and the number of options in the list respectively.

### DIFF
--- a/packages/pf4-component-mapper/src/files/dual-list-select.js
+++ b/packages/pf4-component-mapper/src/files/dual-list-select.js
@@ -145,7 +145,8 @@ const DualList = ({
   sortValues,
   filterValues,
   rightValues,
-  handleValuesClick
+  handleValuesClick,
+  renderStatus
 }) => (
   <FormGroup
     label={label}
@@ -175,6 +176,18 @@ const DualList = ({
                 id={`${input.name}-options-toolbar`}
               />
             </GridItem>
+            {renderStatus && (
+              <GridItem md={12}>
+                <TextContent>
+                  <Text data-test-id="left-status-text" component={TextVariants.h6}>
+                    {renderStatus({
+                      selected: state.selectedLeftValues.length,
+                      options: leftValues.length
+                    })}
+                  </Text>
+                </TextContent>
+              </GridItem>
+            )}
             <GridItem md={12}>
               <List
                 optionClick={handleOptionsClick}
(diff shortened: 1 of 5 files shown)


boaz0:
OK - moved from passing a string to passing a function that gets "props" and returns a string that represents the status text to display below each list search toolbar.
After speaking with @rvsia I realized that it's not necessary to MUI, PF3, BJS and SUIR because each one has its niche so I didn't add it there.
I updated the docs for PF4 to have this props.

//cc @rvsia

boaz0:
Ho yeah - I am going to add tests... :gun:

packages/react-renderer-demo/src/doc-components/examples-texts/pf4/pf4-dual-list.md
```diff
@@ -21,3 +21,4 @@ Dual list select is wrapped in a form group, so it accepts all [form group props
 |filterValueTitle|String|'Filter selected value'|Placeholder for value filter input|
 |filterValueText|String|'Remove your filter to see all selected'|Placeholder for value when there is no filtered value|
 |filterOptionsText|String|'Remove your filter to see all options'|Placeholder for options when there is no filtered option|
+|renderStatus|function|'null'|A function that renders status text below the toolbar filter. For example, display how many items were selected|
```
rvsia, line 24: I would put
```jsx
({selected, options}) => `selected ${selected} from ${options}`
```
 example here

packages/pf4-component-mapper/src/files/dual-list-select.js
```diff
@@ -306,7 +332,8 @@ DualList.defaultProps = {
   filterValueText: 'Remove your filter to see all selected',
   options: [],
   allToLeft: true,
-  allToRight: true
+  allToRight: true,
+  renderStatus: null
```
Hyperkid123, line 336: Can you remove this default prop? It's not necessary.

boaz0: Done :smile_cat:

packages/react-renderer-demo/src/doc-components/examples-texts/pf4/pf4-dual-list.md
boaz0: Done :+1:

packages/pf4-component-mapper/src/files/dual-list-select.js
```diff
@@ -306,7 +332,7 @@ DualList.defaultProps = {
   filterValueText: 'Remove your filter to see all selected',
   options: [],
   allToLeft: true,
-  allToRight: true
+  allToRight: true,
```
Hyperkid123, line 335: I think linter will fail here (trailing comma)

boaz0: dang - you're right!

boaz0: fixed :man_dancing:

boaz0:
Thanks a lot @rvsia @Hyperkid123 - you're amazing!

PR #547, opened 2020-06-09
fix(pf4,suir): Make sort icons consistent among all dual-list selectors

boaz0:
This is a nit change I was asked to do by @nlcwong 

Screenshot_2020-06-09 Data driven forms(1)

Screenshot_2020-06-09 Data driven forms

//cc @rvsia @Hyperkid123

### DIFF
--- a/packages/pf4-component-mapper/src/files/dual-list-select.js
+++ b/packages/pf4-component-mapper/src/files/dual-list-select.js
@@ -18,8 +18,8 @@ import {
 } from '@patternfly/react-core';
 
 import {
-  SortAlphaDownIcon,
-  SortAlphaUpIcon,
+  PficonSortCommonAscIcon,
+  PficonSortCommonDescIcon,
   SearchIcon,
   AngleDoubleLeftIcon,
   AngleDoubleRightIcon,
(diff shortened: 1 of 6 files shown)


rvsia:
@boaz0 Could you please attach Before screenshots? To be honest, I cannot tell the difference. :smile_cat:

boaz0:
Yeah sure!

boaz0:
@rvsia these are the main changes.
In PF4 mapper these icons were used:

image

image

And I changed it to:

image

image

I also noticed that Semantic UI is using arrow icons for sorting:

image

image

and I changed it to:

image

image

I hope that was helpful enough. If you rather me show a diff between the screenshots that's fine too.

rvsia:
@boaz0 oh, this is great! Thanks!

Hyperkid123:
@boaz0 nice. Can you check the layout on smaller screens? I think it's likely that eventually, it will have to be fixed anyway.

boaz0:
@Hyperkid123 is it possible to do it from you laptop using firefox or I need to use my mobile?

Hyperkid123:
@boaz0 just shrink the screen

boaz0:
output

@Hyperkid123

Hyperkid123:
@boaz0 perfect! One more thing I would do is align the buttons next to each other right after the sections stack on each other. I see a small period where the icons point in the right direction but the buttons are still in a column instead of in a row.

boaz0:
output
@Hyperkid123

boaz0:
Thank ya'all @Hyperkid123 @rvsia

PR #553, opened 2020-06-15
Merge ts beta

Hyperkid123:
This should bring the TS changes up to date with the latest PF updates. We can fix up the types from there.


Hyperkid123:
I have removed `@data-driven-forms/react-form-renderer` from the mappers `devDependencies`. It was causing issues with released beta channels and it was installing a different version of the packages to node modules. That caused an invalid final form reference and everything blows up.

There is no need for listing local dependencies (we never released common but we are referencing it in all projects)

PR #554, opened 2020-06-15
fix(pf4): fix wizard types for new version

rvsia:


Hyperkid123:
@rvsia I am close to fixing the issues with the eta channel tags. You should rebase after that.

packages/pf4-component-mapper/src/files/wizard.d.ts
```diff
@@ -28,19 +28,23 @@ export interface WizardNextStepMapper {
...
-  ConditionalNext: React.ComponentType;
-  SubmitButton: React.ComponentType;
-  SimpleNext: React.ComponentType;
-  formOptions: FormOptions;
   disableBack?: boolean;
-  handlePrev: any;
-  nextStep?: string | number;
-  FieldProvider?: React.ComponentType;
-  handleNext: any;
+  handlePrev: Function;
```
Hyperkid123, line 41: So are these provided by the common stuff or does the user actually have to implement these when creating a wizard in the mapper?

rvsia: They are provided.

rvsia: oh :facepalm:

Hyperkid123: Then they should not be listed here. Only the stuff that the developer is required to provide must b listed here.

rvsia: But this is a list of props that are passed to the custom component.. so, should not it be declared to tell developers what they can expect?

```
buttons?: ReactNode | React.ComponentType<WizardButtonsProps>;
```

Hyperkid123: It's not about what developers expect to get. This interface tells what props have to send to `WizardButtons` component. If the `handleNext` is not given the component will not compile.

So the question is, does the dev **have to** pass the handleNext prop to the buttons component? Or is the prop given to the component inside the wizard internally? If it's only handled internally and the developer does not have actually pass it to the component, then it can't be in the interface.

rvsia: The dev have to provide a component that receives handleNext (and other props). It's an internal function passed to a outside component.

Hyperkid123: :+1: Ok then it needs to be there.

PR #562, opened 2020-06-17
PF4 downshift select

### LINKED ISSUE #560: [PF4] Migrate away from react-select
**Scope**: PF4 mapper

**Description**
The react select is now not necessary component because we have all the required logic inside the common code

I would like to use some different libraries (rc-select, downshift, ...) which is lightweight and allows us to customize the component. In an ideal scenario, we would use PF provided component, but that has performance and customization issues that do not allow us to use it at its current state.

Hyperkid123:
closes #560 

Replace react selet with downshift.

Provides complete customization and has a much smaller impact on build size (around 24KB instead of 112KB)

### Current progress

Screenshot from 2020-06-17 16-04-48


Hyperkid123:
The API should remain the same and I've removed the isRead only props. There is no input (by PF design) that can be read-only and there is no point giving that to a div which now holds the value.

Hyperkid123:
> Looks and works great! 
> 
> However, there is one regression - isDisabled on option does not work. You can check the demo wizard schema. isDisabled is being used in sources.

Good catch. I will fix it

PR #564, opened 2020-06-18
Add mobile friendly table of content

rvsia:
On mobile phone the content is put to the end of pages and it's not very useful (especially on long screens):

image

This PR hides this list on smaller screen and puts a menu to the top of the page:

showcontent

+ fix a bug when the code editor was bigger than the viewport.

### DIFF
--- a/packages/react-renderer-demo/src/pages/mappers/blueprint-component-mapper.md
+++ b/packages/react-renderer-demo/src/pages/mappers/blueprint-component-mapper.md
@@ -1,7 +1,10 @@
 import Grid from '@material-ui/core/Grid'
 import ListOfContents from '@docs/list-of-contents';
+import ListOfContentsMobile from '@docs/list-of-contents-select';
 
 <Grid container item>
+
+<ListOfContentsMobile file="mappers/blueprint-component-mapper" />
 <Grid item xs={12} md={10}>
 
 # BlueprintJS
--- a/packages/react-renderer-demo/src/components/code-editor/index.js
+++ b/packages/react-renderer-demo/src/components/code-editor/index.js
@@ -7,6 +7,7 @@ import vsTheme from 'prism-react-renderer/themes/vsDark';
 
 const useStyles = makeStyles({
   pre: {
+    maxWidth: '100vw',
     textAlign: 'left',
     margin: '1em 0',
     padding: '0.5em',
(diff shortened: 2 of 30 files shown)


Hyperkid123:
@rvsia nice
I think we need to emphasize the buttons bit. I think its easy to miss

rvsia:
@Hyperkid123 Yeah, I am thinking the same. However, do you have any idea? icon, make it bold? It should be visible, but not so much. :D

Hyperkid123:
Well, the next question is. Do we even need that in the mobile view? Should we just hide the whole thing?

rvsia:
I think it's a nice feature to have - especially on pages with a lot sections. Users can quickly scan the whole content to find if the page contains what they are searching for. And having it as a small button does not disrupt the look and the feel of the page.

Hyperkid123:
All right let's try to find something that also looks good.

rvsia:
icons

What about? (not pushed yet)

rvsia:
Maybe with pushing it to the right side

Hyperkid123:
I agree it should be on the right side

packages/react-renderer-demo/src/pages/mappers/blueprint-component-mapper.md
```diff
@@ -1,7 +1,10 @@
 import Grid from '@material-ui/core/Grid'
 import ListOfContents from '@docs/list-of-contents';
+import ListOfContentsMobile from '@docs/list-of-contents-select';
 
 <Grid container item>
+
+<ListOfContentsMobile file="mappers/blueprint-component-mapper" />
```
Hyperkid123, line 7: I don't exactly like that we need a special component.
We can use smart layout and reverse the column/row order on small screens. And make this a part of the normal content component.

It should work since we have a different grid item for the actual content and the content navigation.

rvsia: I would like to make a common layout component in the next PR and refactor all pages to use it. I could include this change in it.

```
import DocPage from '@docs/doc-page'

<DocPage file="bla/bla">
  ...content...
</DocPage>
```

instead of defining `Grid`s and contents in every page.

Hyperkid123: Yep, that is exactly what should happen. Can you add a task to the docs project?

rvsia: Done! 🖖 you can find the task in the issues. I will start working on it right after this is merged.

PR #577, opened 2020-06-25
Fix search

rvsia:
- Add zIndex to nav header (it was transparent and the text under it was visible)
- Take URL pathname in nav, not the computed one - errored pages had `_error/` pathnames, not good.
- merge two mappers section


Hyperkid123:
@rvsia we could just give it a background instead of using z-index. Z index can bring other issues.

PR #589, opened 2020-06-26
Add importing section to menu

### LINKED ISSUE #587: [DOCS] Add FAQ section / cjs/umd section
1. useField has to be used in form

- a lot of people encounters this problem and they don't how to deal with that

rvsia:
closes #587

### DIFF
--- a/packages/react-renderer-demo/src/components/navigation/schemas/schema.js
+++ b/packages/react-renderer-demo/src/components/navigation/schemas/schema.js
@@ -61,6 +61,10 @@ const schema = [
     link: 'development-setup'
   },
   {
+    linkText: 'Optimization',
+    link: 'optimization'
+  },
+  {
     linkText: 'Releases',
     link: 'releases'
   },
--- a/packages/react-renderer-demo/src/pages/optimization.md
+++ b/packages/react-renderer-demo/src/pages/optimization.md
@@ -1,0 +1,87 @@
+import DocPage from '@docs/doc-page';
+
+<DocPage>
+
+# Optimization
+
+Data Driven Forms is built in three module formats: CJS, ESM and UMD.
+
+## CJS
+
+```jsx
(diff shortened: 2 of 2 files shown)


packages/react-renderer-demo/src/pages/importing.md
```diff
@@ -0,0 +1,41 @@
...
+
+Data Driven Forms is built in three module formats: UMD, CJS and ESM.
+
+## UMD
+
+```jsx
+import FormRenderer, { useField } from '@data-driven-forms/react-form-renderer';
+```
+
+Only the `FormRenderer` component is imported as a default. In all other packages, use only named exports.
```
Hyperkid123, line 15: this should say exported

packages/react-renderer-demo/src/pages/importing.md
```diff
@@ -0,0 +1,41 @@
+import DocPage from '@docs/doc-page';
+
+<DocPage>
+
+# Importing
+
+Data Driven Forms is built in three module formats: UMD, CJS and ESM.
+
+## UMD
```
Hyperkid123, line 9: I would definitely stress that UMD is there just as a fallback option and should be used if you are building inside CJS or ESM env.

rvsia:
I think calling using of different import as a optimization would be a little bit exaggerated.  :smile_cat:

Hyperkid123:
Yes. It is optimization. The only reason why different builds exist in this project is bundle size optimization. There is no reason to use anything other builds than UMD if you ignore the optimization. For mappers at least. You can reduce the bundle size by a lot by not using components like wizard, select, or date pickers. This is a subject of optimizing applications.

And adding the babel transform example is a great example of how you can automate it. We all sometimes forget to use the correct import. Especially if you are another library (FCE) with multiple builds.

PR #594, opened 2020-06-26
Add env switcher to code examples

rvsia:
Code snippets where are UMD imports of DDF libraries can be switched to CJS/ESM.

switcher


rvsia:
@Hyperkid123 fixed - CJS is now default for all examples

PR #603, opened 2020-07-01
Add fallback to empty notification response

Hyperkid123:
This prevents notifications issues when the response is `null` or empty.

Eventually, we will migrate to something else


rvsia:
Thanks for fixing this!

PR #604, opened 2020-07-01
migrate docs to use firestore

Hyperkid123:
I've created API endpoint for the notifications in firebase functions
No more ambiguous responses we will always receive an array

Also the expiration is simplified. No more start date. Just expiration date. The panel will show everything that has expiration date larger than the current browser date.

### DIFF
--- a/packages/react-renderer-demo/firebaseFunctions.js
+++ b/packages/react-renderer-demo/firebaseFunctions.js
@@ -1,4 +1,5 @@
 const functions = require('firebase-functions');
+const admin = require('firebase-admin');
 const next = require('next');
 const { join } = require('path');
 const { parse } = require('url');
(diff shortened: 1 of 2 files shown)


packages/react-renderer-demo/src/components/notification-panel.js
```diff
@@ -37,38 +37,27 @@ const useStyles = makeStyles((theme) => ({
...
-        return [];
-      }
-
-      if (!Array.isArray(data)) {
-        data = Object.values(data);
-      }
-
-      return data.filter(Boolean).sort((a, b) => b['created-at'].localeCompare(a['created-at']));
-    });
+    .then((data) => data);
```
rvsia, line 42: ^

packages/react-renderer-demo/firebaseFunctions.js
```diff
@@ -13,12 +14,43 @@ const nextjsServer = next({
 });
 const nextjsHandle = nextjsServer.getRequestHandler();
 
+/**
+ * Initialize access to firestore
```
rvsia, line 18: nitpick: I don't think this comment is useful :D

packages/react-renderer-demo/src/components/notification-panel.js
```diff
@@ -37,38 +37,27 @@ const useStyles = makeStyles((theme) => ({
...
+        setNewMessages(data.filter((notification) => !lastSeen.includes(createNotificationId(notification))).length);
+        localStorage.setItem('data-driven-forms-last-seen', JSON.stringify(data.map(createNotificationId)));
+        setNotifications(data);
+      })
+      .catch(() => {
+        setNotifications([]);
+      });
   }, [setNewMessages]);
 
   return (
```
rvsia, line 63: I am sorry, there is actually `created-at` in the old UI code, not `created_at` :-/ So you can change it here.

Hyperkid123: done

PR #606, opened 2020-07-01
Use package title in the main component heading

rvsia:
- testing for algolia wait time
- `{`${avalableMappers.find(({ mapper }) => mapper === activeMapper).title} ${linkText}`}` is because there is a function that creates header id and it need to have only string child

### DIFF
--- a/packages/react-renderer-demo/src/components/component-example-text.js
+++ b/packages/react-renderer-demo/src/components/component-example-text.js
@@ -7,12 +7,12 @@ import avalableMappers from '../helpers/available-mappers';
 const ComponentExampleText = ({ linkText, schema, variants, component, activeMapper, ContentText }) => (
   <React.Fragment>
     <Heading level="4" component="h1">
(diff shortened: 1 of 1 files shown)


packages/react-renderer-demo/src/components/component-example-text.js
```diff
@@ -7,12 +7,12 @@ import avalableMappers from '../helpers/available-mappers';
 const ComponentExampleText = ({ linkText, schema, variants, component, activeMapper, ContentText }) => (
   <React.Fragment>
     <Heading level="4" component="h1">
-      {linkText}
+      {`${avalableMappers.find(({ mapper }) => mapper === activeMapper).title} ${linkText}`}
```
Hyperkid123, line 10: wouldn't this crash before the active mapper property is ready?

rvsia: MUI is always default.

PR #612, opened 2020-07-02
Add tsc config to check d.ts files

rvsia:
`yarn tsc` from the root will check the files.

Currently there are ~~15~~ 2 errors. ~~14~~ ~~2~~ **3** of them are in dependencies (PF4). ~~12 all from material-ui-picker~~.

~~I will try to update MUI pickers and see if that helps.~~ It helped.

**We cannot automate this process because or errors in PF4 package, however, we can at least push it to master and do it manually.**


rvsia:
All errors are solved except: 
- [ ] https://github.com/patternfly/patternfly-react/issues/4467

Hyperkid123 (approved):

PR #633, opened 2020-07-07
fix(ant): refactor and add customization for each component

### LINKED ISSUE #630: [ANT] Check if all components are customizable

rvsia:
closes #630

### DIFF
--- a/packages/ant-component-mapper/src/files/form-template.js
+++ b/packages/ant-component-mapper/src/files/form-template.js
@@ -1,44 +1,82 @@
-/* eslint-disable react/prop-types */
 import React from 'react';
+import PropTypes from 'prop-types';
 import { Button, Typography, Form as AntForm } from 'antd';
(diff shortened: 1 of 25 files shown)


packages/ant-component-mapper/src/files/form-template.js
```diff
@@ -6,7 +6,7 @@ import FormTemplate from '@data-driven-forms/common/src/form-template';
 const { Title, Paragraph } = Typography;
 
 const Form = ({ layout, children, onSubmit, ...props }) => (
-  <AntForm onFinish={onSubmit} layout={layout ? layout : 'vertical'}>
+  <AntForm onFinish={onSubmit} layout={layout ? layout : 'vertical'} {...props}>
```
Hyperkid123, line 9: can we make the `layout` either a default prop or assign it in the function arguments?

Issue #622, opened 2020-07-07
[ANT] Missing components


rvsia:
- [x] FieldArray https://github.com/data-driven-forms/react-forms/pull/637
- [x] Slider #634 
- [x] Duallist #635

PR #635, opened 2020-07-08
fix(ant): add dual list select component

Hyperkid123:
based on: https://ant.design/components/transfer/
cc @Arun-42
part of #622


rvsia:
Dual list implemented by an external library? :thinking:

Hyperkid123:
:fearful:

PR #637, opened 2020-07-09
fix(ant): Added field array component

### LINKED ISSUE #622: [ANT] Missing components

Hyperkid123:
closes #622 
screenshot-localhost_8080-2020 07 09-11_08_53

### DIFF
--- a/packages/ant-component-mapper/src/files/field-array.js
+++ b/packages/ant-component-mapper/src/files/field-array.js
@@ -1,3 +1,209 @@
-import React from 'react';
+import React, { useReducer } from 'react';
+import PropTypes from 'prop-types';
+import { useFieldApi, useFormApi, FieldArray } from '@data-driven-forms/react-form-renderer';
+import { Row, Col, Button, Typography, Space } from 'antd';
+import { UndoOutlined, RedoOutlined } from '@ant-design/icons';
 
-export default () => <div>not implemented</div>;
+import AntForm from '../common/form-wrapper';
+
+const ArrayItem = ({ fields, fieldIndex, name, remove, length, minItems, removeLabel }) => {
+  const { renderForm } = useFormApi();
+
+  const editedFields = fields.map((field, index) => {
+    const computedName = field.name ? `${name}.${field.name}` : name;
+    return { ...field, name: computedName, key: `${computedName}-${index}` };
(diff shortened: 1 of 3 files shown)


packages/ant-component-mapper/src/files/field-array.js
```diff
@@ -1,3 +1,209 @@
...
+          return (
+            <Row gutter={[0, 16]}>
+              <Col span={24}>
+                <Row justify="space-between">
+                  <Col>{label && <Typography.Title level={4}>{label}</Typography.Title>}</Col>
+                  <Col>
+                    <Space>
+                      <Button type="default" onClick={undo} disabled={state.index === 0} icon={<UndoOutlined />} />
+                      <Button type="default" onClick={redo} disabled={state.index === state.history.length} icon={<RedoOutlined />} />
+                      <Button type="primary" onClick={pushWrapper} disabled={value.length >= maxItems}>
```
rvsia, line 144: I think that these buttons could be customizable

rvsia: the same applies for remove button

Hyperkid123: Right, I've completely neglected the customization. I will add it to all of the components.

packages/ant-component-mapper/src/files/field-array.js
```diff
@@ -195,15 +251,46 @@ DynamicArray.propTypes = {
...
+  NoItemsMessageProps: PropTypes.object,
+  ErrorMessageProps: PropTypes.object
 };
 
 DynamicArray.defaultProps = {
   maxItems: Infinity,
   minItems: 0,
   noItemsMessage: 'No items added',
-  FormItemProps: {}
+  FormItemProps: {},
```
rvsia, line 278: As there are no props/classNames merging with these, do we really need all these defaults?

Hyperkid123: There is no difference between `undefined` and `{}` in the spread operators true, but this will help with the docs props generator I have in plan.

PR #648, opened 2020-07-14
feat(pf4): introduce StepTemplate

rvsia:
- component allowing to define custom steps
- also this commit allows to define more props in step definitions

Example:

```jsx
 {
          name: 'step',
          hasNoBodyPadding: true,
          StepTemplate: ({ formFields }) => <div style={{ width: '100%', height: '100%', background: 'red', padding: 32 }}>{formFields}</div>
          ....
}
```

image

cc @PanSpagetka

### DIFF
--- a/packages/pf4-component-mapper/src/files/wizard.d.ts
+++ b/packages/pf4-component-mapper/src/files/wizard.d.ts
@@ -52,6 +52,15 @@ export interface SubstepOfObject {
   title?: ReactNode;
 }
 
+export interface StepTemplateProps {
+  title?: ReactNode;
+  formFields: ReactNode;
+  showTitles?: boolean;
+  showTitle?: boolean;
+  customTitle?: ReactNode;
+  fields: Field[];
+}
+
 export interface WizardField {
   name: string | number;
   fields: Field[];
@@ -62,6 +71,8 @@ export interface WizardField {
   customTitle?: ReactNode;
   disableForwardJumping?: boolean;
   buttons?: ReactNode | React.ComponentType<WizardButtonsProps>;
+  StepTemplate?: React.ComponentType<StepTemplateProps>;
+  hasNoBodyPadding?: boolean;
 }
 
 export interface WizardProps {
@@ -80,6 +91,7 @@ export interface WizardProps {
   closeButtonAriaLabel?: string;
   hasNoBodyPadding?: boolean;
   navAriaLabel?: string;
+  StepTemplate?: React.ComponentType<StepTemplateProps>;
 }
 
 declare const Wizard: React.ComponentType<WizardProps>;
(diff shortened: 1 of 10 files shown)


rvsia:
I skipped the one async wizard tests. We have to deal with it later.

https://github.com/data-driven-forms/react-forms/issues/649

packages/pf4-component-mapper/src/files/wizard.d.ts
```diff
@@ -52,6 +52,15 @@ export interface SubstepOfObject {
   title?: ReactNode;
 }
 
+export interface StepTemplateProps {
+  title?: ReactNode;
+  formFields: ReactNode;
+  showTitles?: boolean;
+  showTitle?: boolean;
```
Hyperkid123, line 59: What is the difference between `showTitle` and `showTitles`? Maybe we could use different names?

packages/pf4-component-mapper/src/files/wizard.js
```diff
@@ -157,7 +159,8 @@ WizardInternal.propTypes = {
   closeButtonAriaLabel: PropTypes.string,
   hasNoBodyPadding: PropTypes.bool,
   navAriaLabel: PropTypes.string,
-  container: PropTypes.instanceOf(Element)
+  container: PropTypes.instanceOf(Element),
+  StepTemplate: PropTypes.func
```
Hyperkid123, line 163: What if somebody provides a class component? I would use the `componentType`

packages/pf4-component-mapper/src/files/wizard.d.ts
rvsia: ShowTitles is on wizard, ShowTitle is on step

packages/pf4-component-mapper/src/files/wizard.js
rvsia: Don't you think elementType?

Hyperkid123: yeah

packages/pf4-component-mapper/src/files/wizard.d.ts
Hyperkid123: Maybe rename the `showTitle` to something like show header? The difference is so tiny that I had to read it 3 times to notice the plural.

rvsia: It doesn't matter. showTitles is set in wizard definition. showTitle in wizard step.

Renaming it would be a breaking change.

Hyperkid123: Ah. Well, that is a bummer. I was not aware it was used already. Nothing we can to at this point.

PR #652, opened 2020-07-15
First commit on new conditions handler

msageryd:
My proposal for a new conditions handler:
- conditions are defined "globally" instead of "locally" per field
- field listeners are registered only for fields involved in "when-clauses"
- the relevant conditions are added as custom data for each field listener (field.data.conditions)
- when conditions triggers, the resulting ui-effects are added to uiState for affected fields via uiStateReducer
- when conditions de-driggers, the ui-effects are removed
- uiState is available in useFormApi(). Any component can check the uiState to decide how/if it should render
- set-instructions in conditions are "one-shot". I.e. when the new field value is set, the set instruction is removed.

TODO:
- TypeScript (I don't know TS, so I'm not the right man to add typing)
- Backwards compability for old conditions structure  (should probably be handled at the top of RegisterConditions).
- My conditionParser differs some from the old one. I put mine in conditions2.js in order to not break other stuff.
- uiStateReducer originally user Immer in my code. I took some shortcuts with the immutability when I removed Immer. This should probably be addressed.
- Documentation
- Decide how to use uiState (hiding and disabling fields etc)
- Circular conditions are possible. It's kind of hard to walk into this problem, but it should probably be handled some way.

I'm sure there are lots to discuss before this reaches production.

### DIFF
--- a/packages/react-form-renderer/src/files/conditions-mapper.js
+++ b/packages/react-form-renderer/src/files/conditions-mapper.js
@@ -1,0 +1,89 @@
+/*
+  conditionsMapper will remap a conditions object and create an object with each depending fieldName as a key.
+
+  Since one field can be involed in more than one condition, an array of condition references will be created under each fieldName key
+
+  Since more than one field can be involved in the same condition, the same condition might be referenced from
+  several condition arrays.
+*/
+
+export const conditionsMapper = ({conditions}) => {
+  if (!conditions) return {};
+
+  function isObject(obj) {
+    return obj !== null && typeof obj === 'object' && !Array.isArray(obj);
+  }
+
+  function isArray(obj) {
+    return Array.isArray(obj);
+  }
+
+  function traverse({obj, fnc, key}) {
+    fnc && fnc({obj, key});
+
+    if (isArray(obj)) {
+      traverseArray({
+        obj,
+        fnc,
+        key,
+      });
+    } else if (isObject(obj)) {
+      traverseObject({
+        obj,
+        fnc,
+        key,
+      });
+    }
+  }
+
+  function traverseArray({obj, fnc, key}) {
+    for (var index = 0, len = obj.length; index < len; index++) {
+      const item = obj[index];
+      traverse({
+        obj: item,
+        fnc,
+        key: index,
+      });
+    }
+  }
+
+  function traverseObject({obj, fnc, key}) {
+    for (var index in obj) {
+      if (obj.hasOwnProperty(index)) {
+        const item = obj[index];
+        traverse({
+          obj: item,
+          fnc,
+          key: index,
+        });
+      }
+    }
+  }
+
+  const indexedConditions = {};
+  const conditionArray = Object.entries(conditions);
+
+  conditionArray
+    .map(([key, condition]) => {
+      return {
+        key: key,
+        ...condition,
+      };
+    })
+    .forEach(condition => {
+      traverse({
+        obj: condition,
+        fnc: ({obj, key}) => {
+          if (key === 'when') {
+            const fieldNames = isArray(obj) ? obj : [obj];
+            fieldNames.map(fieldName => {
+              indexedConditions[fieldName] = indexedConditions[fieldName] || [];
+              indexedConditions[fieldName].push(condition);
+            });
+          }
+        },
+      });
+    });
+
+  return indexedConditions;
+};
--- a/packages/react-form-renderer/src/files/ui-state-reducer.js
+++ b/packages/react-form-renderer/src/files/ui-state-reducer.js
@@ -1,0 +1,111 @@
+const uiStateReducer = (state, action) => {
+  switch (action.type) {
+    case 'conditionResult': {
+      const {
+        source,
+        uiState: {add, remove},
+      } = action;
+
+      const validAttributes = [
+        'visible',
+        'disabled',
+        'hidden',
+        'enabled',
+        'icon',
+        'help',
+        'colorBar',
+        'required',
+        'max',
+        'min',
+      ];
+
+      //enabled and hidden are not used internally, but accepted as input from the conditions object
+      const inversedTypes = {
+        enabled: 'disabled',
+        hidden: 'visible',
+      };
+
+      const removeItem = state => {
+        if (!remove) return;
+
+        Object.entries(remove).forEach(([key, item]) => {
+          //If the field/section doesn't exist, we don't need to do anymore
+          if (!state[key]) return;
+
+          validAttributes.forEach(type => {
+            //Don't process uiTypes if they don't exist in the dispatched message
+            if (item[type] === undefined) return;
+
+            const inversedType = inversedTypes[type];
+            type = inversedType || type;
+
+            if (remove[key][type]) {
+              const index = state[key][type].findIndex(item => item.source === source);
+              if (index !== -1) state[key][type].splice(index, 1);
+
+              //If this was the last item of this type, remove the type
+              if (state[key][type].length === 0) delete state[key][type];
+            }
+          });
+
+          //If no more uiStateType keys exists for this field, remove the field
+          if (Object.keys(state[key]).length === 0) delete state[key];
+        });
+      };
+
+      const addItem = state => {
+        if (!add) return;
+
+        Object.entries(add).forEach(([key, item]) => {
+          //Create the item object for this item if it doesn't exist
+          if (!state[key]) state[key] = {};
+
+          validAttributes.forEach(type => {
+            //Don't add uiTypes if they don't exist in the dispatched message
+            if (item[type] === undefined) return;
+
+            //Handle inversed types (disabled/enabled, visible/hidden)
+            const inversedType = inversedTypes[type];
+            const value = inversedType ? !item[type] : item[type];
+            type = inversedType || type;
+
+            if (!state[key][type]) {
+              //If this type doesn't exists for this item, we create a new array with only this source. No need to search fot the source
+              state[key][type] = [{source, value}];
+            } else {
+              const index = state[key][type].findIndex(item => item.source === source);
+              if (index !== -1) {
+                //If this type for this item from this source existed, update the state (could change if condition went from "then" to "else")
+                state[key][type][index].value = value;
+              } else {
+                //Otherwise, add the state from this source at the begining of the array (i.e. this will supress result from other sources)
+                state[key][type].unshift({source, value});
+              }
+            }
+          });
+
+          //Set-instructions are ephemeral and goes into a separate list which is emptied when processed
+          if (item.set) {
+            state.setFieldValues = {...state.fileldValues, [key]: item.set};
+          }
+        });
+      };
+
+      let mutatedState = state;
+      if (remove) {
+        removeItem(mutatedState);
+      }
+      //If uiStates should be added, go through all add fields and all possible types for these fields
+      if (add) {
+        addItem(mutatedState);
+      }
+      console.log(mutatedState);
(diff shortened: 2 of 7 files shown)


Hyperkid123:
So the rendering performance seems solid. Looking at the profiler, the only time when multiple fields are re-rendered in the same cycle is when a condition is met/broken which makes sense and has to happen. :+1: on that.

I will now dig a bit deeper trough the code.

packages/react-form-renderer/src/files/conditions-mapper.js
```diff
@@ -0,0 +1,89 @@
...
+        obj: item,
+        fnc,
+        key: index,
+      });
+    }
+  }
+
+  function traverseObject({obj, fnc, key}) {
+    for (var index in obj) {
+      if (obj.hasOwnProperty(index)) {
```
Hyperkid123, line 52: This check is redundant.  You are checking if an index from a set of object indexes is present inside the object.

You can use `Object.entries` in your traverse functions instead of the `for in` loop. That will give you both the index and value as params.

packages/react-form-renderer/src/files/ui-state-reducer.js
```diff
@@ -0,0 +1,92 @@
...
+};
+
+const uiStateReducer = (state, action) => {
+  switch (action.type) {
+    case 'addUIState': {
+      const {source, uiState} = action;
+
+      Object.entries(uiState).forEach(([key, item]) => {
+        //Create the item object for this item if it doesn't exist
+        if (!state.fields[key]) state.fields[key] = {};
```
Hyperkid123, line 27: you should probably create a new copy state object before mutating the variable. We might encounter issues with async rendering and invalid state.

packages/react-form-renderer/src/files/ui-state-reducer.js
```diff
@@ -0,0 +1,92 @@
...
+            //If this type doesn't exists for this item, we create a new array with only this source. No need to search fot the source
+            state.fields[key][type] = [{source, value}];
+          } else {
+            const index = state.fields[key][type].findIndex(item => item.source === source);
+            if (index !== -1) {
+              //If this type for this item from this source existed, update the state (value could change if condition went from "then" to "else")
+              state.fields[key][type][index].value = value;
+            } else {
+              //Otherwise, add the state from this source at the begining of the array (i.e. this will supress result from other sources)
+              state.fields[key][type].unshift({source, value});
```
Hyperkid123, line 48: the `{...foo}` is not a deep clone. If we want to change nested non-primitive types we have to clone them as well. Using `push`, `shift` etc. will not create a new instance and that might not trigger React render function. For example:

```jsx
x = [1,2,3]
a = {b: {c: x}}
b = {...a}
a.b.c === b.b.c // results in true and that will not trigger new render
```

packages/react-form-renderer/src/files/conditions-mapper.js
msageryd: I'll change to `entries` for better readability. But I do believe that the check for hasOwnProperty is needed in order to ensure that no inherited properties are traversed. This should not be the case with your schema object though.

msageryd: Fixed in 21361ec. I'm not sure how I should "resolve" these comments. Should I somehow link them to the commnit?

packages/react-form-renderer/src/files/conditions-mapper.js
```diff
@@ -0,0 +1,89 @@
...
+  Since one field can be involed in more than one condition, an array of condition references will be created under each fieldName key
+
+  Since more than one field can be involved in the same condition, the same condition might be referenced from
+  several condition arrays.
+*/
+
+export const conditionsMapper = ({conditions}) => {
+  if (!conditions) return {};
+
+  function isObject(obj) {
```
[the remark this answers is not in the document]
msageryd: Fixed in 21361ec.

packages/react-form-renderer/src/files/ui-state-reducer.js
msageryd: Fixed in 12ce210. I'm not used to handle immutability manually in plain JS. Please check the code so I got it right.

packages/react-form-renderer/src/files/ui-state-reducer.js
msageryd: Fixed in 12ce210

msageryd:
DONE:
- *uiStateReducer originally user Immer in my code. I took some shortcuts with the immutability when I removed Immer. This should probably be addressed.*
**Fixed in commit 12ce210**

- *Circular conditions are possible. It's kind of hard to walk into this problem, but it should probably be handled some way.*
**I cannot reproduce my circular scenario anymore. I'm guessing that this problem is solved due to the better immutability handling in uiStateReducer.**

msageryd:
DONE (98d2064):
- *My conditionParser differs some from the old one. I put mine in conditions2.js in order to not break other stuff.*
**conditions2.js has now replaced the old conditions.js**

- *Decide how to use uiState (hiding and disabling fields etc)*
**I decided that FormFieldConditionWrapper should be removed. The condition data is already available in uiState in context and used as input to FormFieldHideWrapper**

msageryd:
@Hyperkid123 Some design questions:

**Do you want the conditions collection to be an object or an array?**

Object pros: 
- Forces user to define names for each condition. 
- Easier to debug with good names.
- Easier to merge multiple conditions objects

Array pros: 
- Persisted order of conditions

**Is support for sequence still needed?**

You said earlier that you wanted to get rid of sequence conditions. Does this mean that the legacyConditionsMapper does not have to cover sequences?

msageryd:
@Hyperkid123 I can't find any repo for the documentation. Thought I'd update the docs for the new conditions schema.

msageryd:
@Hyperkid123 I've implemented a naive legacyConditionsMapper. It can only handle simple visibility conditions for now. This needs to be revisited in order to handle more complex conditions. f1818eb

msageryd:
@Hyperkid123 I have started to update the tests for conditions. How do I run the tests? `yarn test` seems to run the tests, but all test fails, so I suppose I need to do something more.

Hyperkid123:
@msageryd sorry for the silence (weekend). Going through the changes now.

Hyperkid123:
> I decided that FormFieldConditionWrapper should be removed. The condition data is already available in uiState in context and used as input to FormFieldHideWrapper

This might cause issues with the value initialization and submission. There is a difference between the condition wrapper and the hidden wrapper.

If the **condition** is not met the field **is not in the DOM** and once the condition is met **you can trigger mounting events** like re-initialize the field. So if the condition is false the **value is not sent to the submit callback**. Yes value of the condition can be persisted even if it's hidden but that is aa opt-in attribute.

If the **hideField** is `true`, the filed is not visible but **still present in the DOM** and its **value is submitted**. It is basically an input of type hidden for non-input fields.

So the condition wrapper might no longer be required but we still need to detach/attach the field from/to the DOM.

Hyperkid123:
> **Do you want the conditions collection to be an object or an array?**

From the development point of view, the object is preferable. But the order of execution is important and it makes it a bit easier for a user to understand and define the condition order. That is ultimately the reason why the whole schema is a nested array since its sometimes created by users who do not necessarily have any experience with any development. I would like to stick with the Array if possible.

> **Is support for sequence still needed?**
> 
> You said earlier that you wanted to get rid of sequence conditions. Does this mean that the legacyConditionsMapper does not have to cover sequences?

Sequences still have to be supported. We have to keep them at least until the next major release. We introduced them to ensure the correct order of execution of conditions using **AND** and **OR** statements when we introduced the `set` outcome of a condition.

Hyperkid123:
> @Hyperkid123 I can't find any repo for the documentation. Thought I'd update the docs for the new conditions schema.

its the `react-renderer-demo` package (I know what an intuitive name for a docs package). https://data-driven-forms.org/development-setup#rundocumentation

I would like to add that if run locally, the docs use locally build packages so it is required that all packages are built on your local machine. Running `yarn build` in the root directly would do that for you.

Hyperkid123:
> @Hyperkid123 I have started to update the tests for conditions. How do I run the tests? `yarn test` seems to run the tests, but all test fails, so I suppose I need to do something more.

There is the same requirement as with the docs. All packages must be built locally in since the tests use a local build of all packages. This is there to ensure that no matter what changes in what packages you make, it will run all the tests with all of these changes. If you have all packages build already and you still see some error, just post the stack trace here and we can look at it.

But looking at the circle CI log, you have a circular dependency so the renderer package wont even build:
```
./src/index.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/component-types.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/compose-validators.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/conditions-mapper.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/data-types.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/default-schema-validator.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/field-array.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/field-provider.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/form-renderer.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/form-spy.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/form.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/legacy-conditions.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/register-conditions.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/renderer-context.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/schema-errors.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/set-field-values.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/ui-state-reducer.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/use-field-api.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/use-form-api.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/validator-mapper.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/validator-types.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/validators.js, /home/circleci/react-forms/packages/react-form-renderer/src/files/wizard-context.js → ./dist/cjs...
 
[!] Error: Circular dependency: src/index.js -> src/files/form-renderer.js -> src/files/register-conditions.js -> src/index.js
 
Error: Circular dependency: src/index.js -> src/files/form-renderer.js -> src/files/register-conditions.js -> src/index.js
 
    at Object.onwarn (/home/circleci/react-forms/packages/react-form-renderer/rollup.config.js:43:9)
 
    at Object.config.onwarn.warning [as onwarn] (/home/circleci/react-forms/node_modules/rollup/dist/shared/node-entry.js:13599:25)
 
    at Graph.config.onwarn.warning [as onwarn] (/home/circleci/react-forms/node_modules/rollup/dist/shared/node-entry.js:13599:25)
 
    at Graph.warn (/home/circleci/react-forms/node_modules/rollup/dist/shared/node-entry.js:13403:14)
 
    at Graph.link (/home/circleci/react-forms/node_modules/rollup/dist/shared/node-entry.js:13420:18)
 
    at Promise.all.then (/home/circleci/react-forms/node_modules/rollup/dist/shared/node-entry.js:13286:18)
 
    at process._tickCallback (internal/process/next_tick.js:68:7)
 
```

That may be the reason why the tests fail.

PR #653, opened 2020-07-16
Use pf4 select component

Hyperkid123:
This just ate my whole day!

### Changes
- use typeahead styling for PF4 select
- add top cropping of the menu if the element is of the scrollable parent element (bottom is not needed since the the menu should overflow the parent)
- add scroll overflow to the menu to prevent window overflow

zkurveny-select

### DIFF
--- a/packages/pf4-component-mapper/demo/index.js
+++ b/packages/pf4-component-mapper/demo/index.js
@@ -24,7 +24,7 @@ const fieldArrayState = { schema: arraySchemaDDF, additionalOptions: {
 class App extends React.Component {
     constructor(props) {
         super(props);
-        this.state = {schema: wizardSchema, additionalOptions: {}} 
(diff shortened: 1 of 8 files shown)


packages/pf4-component-mapper/demo/index.js
```diff
@@ -53,25 +53,42 @@ class App extends React.Component {
...
-                    },
-                    summary: Summary
-                }}
-                FormTemplate={(props) => <FormTemplate {...props} showFormControls={this.state.additionalOptions.showFormControls} />}
-                onCancel={console.log}
-                schema={this.state.schema}
-                uiSchema={this.state.ui}
-                {...this.state.additionalOptions}
-            />
+            <Modal isOpen variant="small" title="title">
```
rvsia, line 56: Can you please return the demo to normal version? :D

Hyperkid123:
done

Issue #650, opened 2020-07-14
PF3 datepicker tests fail outside our timezone


rvsia:
**Scope**: PF3

**Description**

Set India's timezone (or any other that has plus some minutes) and run tests for PF3 date-time-picker

`react-forms/packages/pf3-component-mapper/src/tests/form-fields/date-time-picker/time-picker.test.js`

Two tests are going to fail.

@Arun-42 Do you want to take this? I think it could be interesting entry point to tests.

Hyperkid123:
I wonder if its possible to set the JSDOM navigator to one locale in tests :thinking:. It might be read-only property.

Arun-42:
Yes, I want to take this issue.

PR #654, opened 2020-07-16
PF3 datepicker test timezone fix

Arun-42:
Sets the global jest environment timezone to UTC.
Fix: #650

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,8 @@
     },
     "modulePathIgnorePatterns": [
       "<rootDir>/templates/"
-    ]
+    ],
+    "globalSetup": "<rootDir>/globalSetup.js"
   },
   "devDependencies": {
     "@babel/plugin-transform-runtime": "^7.9.0",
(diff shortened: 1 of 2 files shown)


Arun-42:
Copied from https://github.com/jest-community/vscode-jest/issues/153#issuecomment-543814787.

package.json
```diff
@@ -40,7 +40,8 @@
     },
     "modulePathIgnorePatterns": [
       "<rootDir>/templates/"
-    ]
+    ],
+    "globalSetup": "<rootDir>/globalSetup.js"
```
rvsia, line 44: I don’t know what the difference is 🤔  but have you tried to setup it in `jest.setup.js`?

Arun-42: From [here](https://jestjs.io/docs/en/configuration), I see the difference is that `globalSetup` is run once at the beginning while `setupFiles` is run before each test. But setting `process.env.TZ` in `jest.setup.js` isn't working. I am not sure why :thinking:.

PR #657, opened 2020-07-21
feat(manager): create subscription based state manager

Hyperkid123:
### Changes
Added initial implementation of state manager context, `useField` and `useSubscriotion` hooks. This will render fields on demand not on every state change.
- fields are currently only subscribed to their state mutation
- form context only extracts field values on form submit
  - we are using class instance (closure) per field to hold its internal state
  - class method reference is registered on the field mount to context and called on-demand to retrieve the field state (currently only on submit)
  - the instance must be synchronized when the field state changes

#### Example of independent rendering using the profiler tool:
state-manager-subscription

Note: I am thinking about writing this purely in TS. We should make the decision ASAP so it won't be much work if we decide to do it later. TS will help a lot with all of these interfaces and functions.

### DIFF
--- a/packages/form-state-manager/src/utils/state-manager-reducer.js
+++ b/packages/form-state-manager/src/utils/state-manager-reducer.js
@@ -1,0 +1,69 @@
+export const REGISTER_FIELD = 'REGISTER_FIELD';
+export const UNREGISTER_FIELD = 'UNREGISTER_FIELD';
+
+const registerField = (state, { name, getFieldState }) => ({
+  ...state,
+  registeredFields: [...state.registeredFields, name],
+  fieldListeners: {
+    ...state.fieldListeners,
+    [name]: {
+      /**
+       * There can be multiple listeners checking the same field
+       * We need to check if we can destroy all listeners before we remove all of them
+       */
+      count: state.fieldListeners?.[name]?.count + 1 || 1,
+      getFieldState
+    }
+  }
+});
+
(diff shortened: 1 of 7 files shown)


packages/form-state-manager/src/utils/state-manager-reducer.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+});
+
+const checkLastSubscrition = (fieldListeners, name) => fieldListeners?.[name]?.count === 1;
+
+const unregisterField = (state, { name, persistOnUnmount, value }) => {
+  const isLast = checkLastSubscrition(state.fieldListeners, name);
+  if (!isLast) {
+    return state;
```
rvsia, line 25: Should not the count be decreased by one?

Hyperkid123: Yes it should!

packages/form-state-manager/src/utils/state-manager-reducer.js
```diff
@@ -0,0 +1,69 @@
...
+    return {
+      ...state,
+      fieldListeners: {
+        ...state.fieldListeners,
+        [name]: {
+          ...state.fieldListeners[name],
+          count: state.fieldListeners[name].count - 1
+        }
+      }
+    };
```
rvsia, line 34: just nitpick: I think this could be extracted to a separate function for better readability.

PR #658, opened 2020-07-21
feat(manager): convert files to TS

Hyperkid123:
TODO
- [x] convert current files to TS
- [x] set up build
- [x] set up tests


packages/form-state-manager/src/utils/use-subscription.ts
```diff
@@ -1,35 +1,46 @@
...
 
-    return event.target.type === 'checkbox' ? event.target.checked : event.target.value;
+    return event?.target.type === 'checkbox' ? event.target.checked : event?.target.value;
   }
 
   return event;
 };
 
-const useSubscription = ({ name, initialValue, subscription = {} }) => {
+interface UseSubscriotion {
```
rvsia, line 37: typo

Hyperkid123:
@rvsia I had finally some spare time to move this forwards. Fortunately, we can use existing test configuration so we are good to go. I can now focus a create some more tasks to actually add features to the state manager and subscription.

PR #659, opened 2020-07-21
fix(pf4): allow more customization

rvsia:
- introduces FormGroupProps
- validated can be overwritten by props
- some types and props bugs fixed

example

```jsx
{
    component: 'text-field',
    name: 'validate',
    label: 'Async validation',
    validate: [asyncValidate, {
        type: 'required'
    }],
    resolveProps: (input, {
        meta
    }) => {
        console.log(meta.validating)

        if (meta.validating) {
            return {
                helperText: <span><Spinner size="md" />Validating</span>
            }
        }

        if (meta.valid) {
            return {
                helperText: "It's okay!",
                validated: 'success',
                FormGroupProps: {
                    validated: 'success'
                }
            }
        }

        return {}
    }
}
```

poc

### DIFF
--- a/packages/pf4-component-mapper/src/common/form-group.js
+++ b/packages/pf4-component-mapper/src/common/form-group.js
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 
 import showError from './show-error';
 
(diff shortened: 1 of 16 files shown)


Hyperkid123:
@rvsia there are just some linting issues.

PR #671, opened 2020-08-06
feat(manager): create global manager api

### LINKED ISSUE #669: [manager] Create global state manager instance
- it is detached from react rendering cycle
- only serves as an API to get the state data
- does not update the form context at first
  - that will be handled by subscription later on
- add basic API
  - store values
  - change form value
  - registered fields list

Hyperkid123:
closes #669 

- [x] add tests

### DIFF
--- a/packages/form-state-manager/src/files/form-manager-context.ts
+++ b/packages/form-state-manager/src/files/form-manager-context.ts
@@ -13,6 +13,7 @@ const FormManagerContext = createContext<ManagerContextValue>({
   handleSubmit: noop,
   registerField: noop,
   unRegisterField: noop,
+  change: noop
 });
 
 export default FormManagerContext;
--- a/packages/form-state-manager/src/files/form-state-manager.tsx
+++ b/packages/form-state-manager/src/files/form-state-manager.tsx
@@ -1,6 +1,7 @@
-import React, { useReducer, FormEvent } from 'react';
+import React, { useReducer, FormEvent, useRef } from 'react';
 
 import FormManagerContext from './form-manager-context';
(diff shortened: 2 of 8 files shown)


packages/form-state-manager/src/files/form-manager-context.ts
```diff
@@ -13,6 +13,7 @@ const FormManagerContext = createContext<ManagerContextValue>({
   handleSubmit: noop,
   registerField: noop,
   unRegisterField: noop,
```
rvsia, line 15: It does not belong to this PR... but unregister should be one word `unregisterField`. :smile:

Hyperkid123:
looks like npm is down :question: CI is failing to install dependencies
```
error An unexpected error occurred: "https://registry.yarnpkg.com/@khala/npm-release-monorepo/-/npm-release-monorepo-2.4.1.tgz: Request failed \"404 Not Found\"".
```

packages/form-state-manager/src/files/form-manager-context.ts
Hyperkid123: yup I think this whole state will move to the global manager API. And the context will be eventually updated only when subscribed.

Hyperkid123:
now it fails on: 
```
error An unexpected error occurred: "https://registry.yarnpkg.com/@definitelytyped/utils/-/utils-0.0.34.tgz: Request failed \"404 Not Found\"".
```

I will restart it in a couple of minutes. Locally it installs fine...

Hyperkid123:
@rvsia you can merge it if you want everything is working fine locally and this does not change any other package... IF we want to continue work on this today we will need this PR.

PR #674, opened 2020-08-06
feat(state): move context to manager api

rvsia:


Hyperkid123:
I've run the tests a build locally. Npm is still having issues with scoped packages.

PR #678, opened 2020-08-06
feat(state): Add basic form metadata

### LINKED ISSUE #675: [manager] Add form metadata

rvsia:
closes #675

Add all metadata.

Added computation example for change field.

### DIFF
--- a/packages/form-state-manager/src/utils/manager-api.ts
+++ b/packages/form-state-manager/src/utils/manager-api.ts
@@ -18,10 +18,38 @@ const createManagerApi: CreateManagerApi = (onSubmit) => {
     unregisterField,
     getState,
     registeredFields: [],
-    fieldListeners: {}
+    fieldListeners: {},
+    active: null,
+    dirty: false,
+    dirtyFields: [],
+    dirtyFieldsSinceLastSubmit: [],
+    dirtySinceLastSubmit: false,
+    error: null,
+    hasSubmitErrors: false,
+    hasValidationErrors: false,
+    initialValues: {},
+    invalid: false,
+    modified: {},
+    modifiedSinceLastSubmit: false,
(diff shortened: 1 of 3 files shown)


packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -18,10 +18,38 @@ const createManagerApi: CreateManagerApi = (onSubmit) => {
...
+    visited: {}
   };
+
   function change(name: string, value?: any): any {
     state.values[name] = value;
+    state.visited[name] = true;
+    state.modified[name] = true;
+    state.modifiedSinceLastSubmit = true;
+    state.dirtySinceLastSubmit = true;
+    state.dirtyFields.push(name);
```
Hyperkid123, line 51: This will add the same name multiple times to the array on each change. I don't think we want that. We can easily check if the field is pristine/dirty and make a set of initial changes only once. (pristine, dirty, touched, etc.)

rvsia: You are right.

But I feel that all this computation will be refactored a lot of times since now, so this just a 'placeholder'.

But it's useful to write tests, because they won't be changed.

packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -7,6 +7,8 @@ import FieldState from '../types/field-state';
 
 const isLast = (fieldListeners: AnyObject, name: string) => fieldListeners?.[name]?.count === 1;
 
+const addIfUnique = (array: Array<any>, item: any) => !array.includes(item) && array.push(item);
```
Hyperkid123, line 10: I feel like this should be `Array<string>`. We should try to use `any` as little as possible.

PR #682, opened 2020-08-06
feat(state): add meta for fields

### LINKED ISSUE #680: [manager] Add field meta data

rvsia:
closes #680


packages/form-state-manager/src/types/field-config.ts
```diff
@@ -0,0 +1,10 @@
+import AnyObject from './any-object';
```
Hyperkid123, line 1: this file should have `.d.ts` extension. Seems like the old one had the wrong extension as well.

rvsia:
@Hyperkid123 CircleCI seems to be ill :mask:

PR #689, opened 2020-08-07
feat(state): introduce initializeOnMount

### LINKED ISSUE #688: [manager] Implement initiliazeOnMount
- field level
- form leve

rvsia:
closes #688

### DIFF
--- a/packages/form-state-manager/src/tests/utils/manager-api.test.js
+++ b/packages/form-state-manager/src/tests/utils/manager-api.test.js
@@ -180,4 +180,42 @@ describe('managerApi', () => {
     blur('foo');
     expect(managerApi().active).toBeUndefined();
   });
+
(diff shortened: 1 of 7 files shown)


packages/form-state-manager/src/tests/utils/manager-api.test.js
```diff
@@ -180,4 +180,42 @@ describe('managerApi', () => {
     blur('foo');
     expect(managerApi().active).toBeUndefined();
   });
+
+  describe('initializeOnMount', () => {
+    it.skip('should initialize on form level using initialValues', () => {
```
Hyperkid123, line 185: skip :hole:

rvsia: intentional, we are not supporting initialValues yet

Issue #687, opened 2020-08-07
[manager] Add validation support


Hyperkid123:
- [x] field-level sync validation #691 
- [x] field-level async validation #691
- [x] form-level sync validation @rvsia 
- [x] form-level async validation @rvsia 
- [x] mutate field-level validation based state attributes https://github.com/data-driven-forms/react-forms/pull/700
- mutate form-level validation based state attributes
  - [x] field level https://github.com/data-driven-forms/react-forms/pull/703
  - [x] form level https://github.com/data-driven-forms/react-forms/pull/719
  - [x] sync form and field level validation output https://github.com/data-driven-forms/react-forms/pull/719
- [x] field-level multi-async validation support https://github.com/data-driven-forms/react-forms/pull/715

PR #691, opened 2020-08-07
feat(manager): add field level validation API

Hyperkid123:
part of #687 

### Changes
- adds managerApi validation flag mutation
- adds field level sync/async validator

### DIFF
--- a/packages/form-state-manager/src/utils/validate.ts
+++ b/packages/form-state-manager/src/utils/validate.ts
@@ -1,0 +1,17 @@
+import AnyObject from '../types/any-object';
+import { FieldLevelValidator } from '../types/validate';
+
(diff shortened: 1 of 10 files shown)


packages/form-state-manager/src/utils/validate.ts
```diff
@@ -0,0 +1,23 @@
...
+     * more info: https://github.com/microsoft/TypeScript/issues/29642
+     */
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const asyncResult = result as Promise<string | undefined>;
+    managerApi().registerAsyncValidator(asyncResult);
+    asyncResult.then(() => undefined).catch((error) => error);
+    return result;
+  }
+
+  return validator(value, allValues);
```
rvsia, line 22: return result

PR #694, opened 2020-08-10
feat(state): add subscription API

### LINKED ISSUE #693: [manager] Create a subscription API

rvsia:
closes #693

### DIFF
--- a/packages/form-state-manager/src/utils/manager-api.ts
+++ b/packages/form-state-manager/src/utils/manager-api.ts
@@ -14,6 +14,7 @@ const shouldExecute = (formLevel: boolean | undefined, fieldLevel: boolean | und
 
 type objectMapFunction = (value: any, key: any) => any;
 
+// TODO: try to optimize
 const traverseObject = (object: AnyObject, callback: objectMapFunction) => Object.keys(object).forEach((key) => callback(object[key], key));
 
 const asyncWatcher: AsyncWatcher = (updateValidating, updateSubmitting) => {
--- a/packages/form-state-manager/src/utils/manager-api.ts
+++ b/packages/form-state-manager/src/utils/manager-api.ts
@@ -1,16 +1,23 @@
 import { FormEvent } from 'react';
 import set from 'lodash/set';
 
-import CreateManagerApi, { ManagerState, ManagerApi, AsyncWatcher, AsyncWatcherRecord } from '../types/manager-api';
+import CreateManagerApi, { ManagerState, ManagerApi, AsyncWatcher, AsyncWatcherRecord, Rerender } from '../types/manager-api';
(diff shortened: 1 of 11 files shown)


packages/form-state-manager/src/utils/use-subscription.ts
```diff
@@ -54,16 +56,18 @@ export const initialMeta = (initial: any): Meta => ({
...
     name,
-    meta: initialMeta(initialValue)
-  });
+    meta: initialMeta(initialValue),
+    internalId: generateId()
+  }));
   const {
     current: { getDetachedState, setDetachedState }
   } = useRef(createFieldState(state));
+  const [_, render] = useReducer((count) => count + 1, 0);
```
Hyperkid123, line 70: Any specific reason why using reducer and not a state?

also `const [, render] => ....` should bee good enough. No need for the underscore.

packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -1,17 +1,24 @@
...
+
 const isLast = (fieldListeners: AnyObject, name: string) => fieldListeners?.[name]?.count === 1;
 
 const addIfUnique = (array: Array<string>, item: string) => !array.includes(item) && array.push(item);
 
 const shouldExecute = (formLevel: boolean | undefined, fieldLevel: boolean | undefined) => (formLevel || fieldLevel) && fieldLevel !== false;
 
+type objectMapFunction = (value: any, key: any) => any;
+
+const traverseObject = (object: AnyObject, callback: objectMapFunction) => Object.keys(object).forEach((key) => callback(object[key], key));
```
Hyperkid123, line 20: Any way to memoize this function? It could be the most expensive operation.

packages/form-state-manager/src/utils/use-subscription.ts
rvsia: With `useState` you will have to still write `render((prev) => prev + 1))` so this seems to be a little bit cleaner in my eyes. Not sure what about performance, but Redux is using the same construction.

packages/form-state-manager/src/utils/manager-api.ts
rvsia: I would keep optimization as the last step in our building. (We can then do some performance testing.)

Hyperkid123: Can you add a TODO to the function? We will forget it. (I will)

rvsia: I did it. However, I have seen so many TODO comments I become immune to them. :smiley: 

Yeah, but we can focus on performance after we finalize the whole state construction. There are still so many blank spaces right now.

PR #697, opened 2020-08-10
fix(pf4): Add customization api to plain text

Hyperkid123:


packages/pf4-component-mapper/src/files/plain-text.js
```diff
@@ -4,11 +4,11 @@ import PropTypes from 'prop-types';
 import { TextContent, Text } from '@patternfly/react-core';
 import validTextFields from '@data-driven-forms/common/src/utils/valid-text-fields';
 
-const PlainText = ({ label, name, variant }) => (
-  <TextContent>
+const PlainText = ({ label, name, variant, TextProps, TextContentProps }) => (
```
rvsia, line 7: I think `TextProps` should be changed to `...rest` to make it consistent with the way we are spreading props in all other components. (`...rest` should be always spread to the main component.)

Just catch `component` props before you do it.

packages/pf4-component-mapper/src/files/plain-text.d.ts
```diff
@@ -1,9 +1,12 @@
 import { ReactNode } from "react";
+import { TextProps, TextContentProps } from "@patternfly/react-core";
 
 export interface PlainTextProps {
   label: ReactNode;
   name: string;
   variant?: 'p'|'span'|'strong'|'b'|'cite'|'caption'|'code'|'em'|'i'|'h1'|'h2'|'h3'|'h4'|'h5'|'h6'|'h6'|'div'|'label'|'pre'|'q'|'samp'|'small'|'sub'|'sup';
+  TextProps: TextProps,
+  TextContentProps: TextContentProps,
```
rvsia, line 9: Maybe we should add it do docs.

Hyperkid123: Ah right its not generated yet.

packages/pf4-component-mapper/src/files/plain-text.js
```diff
@@ -4,11 +4,11 @@ import PropTypes from 'prop-types';
 import { TextContent, Text } from '@patternfly/react-core';
 import validTextFields from '@data-driven-forms/common/src/utils/valid-text-fields';
 
-const PlainText = ({ label, name, variant }) => (
-  <TextContent>
+const PlainText = ({ label, name, variant, TextContentProps, ...rest }) => (
```
rvsia, line 7: please catch 'component' prop coming from renderer, it could cause issues.

packages/pf4-component-mapper/src/files/plain-text.js
```diff
@@ -19,7 +19,8 @@ const PlainText = ({ label, name, variant }) => (
 PlainText.propTypes = {
   variant: PropTypes.oneOf(validTextFields),
```
rvsia, line 20: We should mention that `component` PF prop is mapped to `variant` prop in Docs.

packages/pf4-component-mapper/src/files/plain-text.js
Hyperkid123: I think this actually is closer to the current way of customization in our other mappers (stuff like required, disabled, etc.). Plus the component prop is always caught by the use field hook. And you can't really use the `component` key in the schema.

rvsia: The plaintext component will receive `component="plain_text"` and this is then passed via `rest` to the PF `Text` component. There is no `useFieldApi` hook here to catch it.

Hyperkid123: we do not catch `component` in useFIeldApi!?

rvsia: We do, but plain text is not using it. It's not a field, it's just a component.

PR #698, opened 2020-08-10
feat(state): implement clearedValue

### LINKED ISSUE #696: [manager] Implement clearedValue
- field level
- form level

rvsia:
closes #696

### DIFF
--- a/packages/form-state-manager/src/tests/utils/use-subscription.test.js
+++ b/packages/form-state-manager/src/tests/utils/use-subscription.test.js
@@ -3,7 +3,7 @@ import React, { useEffect } from 'react';
 import { act } from 'react-dom/test-utils';
 import { mount } from 'enzyme';
 import FormManagerContext from '../../files/form-manager-context';
-import useSubscription, { initialMeta } from '../../utils/use-subscription';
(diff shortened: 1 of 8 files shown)


packages/form-state-manager/src/tests/utils/use-subscription.test.js
```diff
@@ -211,4 +211,77 @@ describe('useSubscription', () => {
...
+  });
+
+  it('#checkEmpty - edge cases', () => {
+    expect(checkEmpty(false)).toEqual(false);
+    expect(checkEmpty('A')).toEqual(false);
+    expect(checkEmpty(0)).toEqual(false);
+
+    expect(checkEmpty([])).toEqual(true);
+    expect(checkEmpty('')).toEqual(true);
+    expect(checkEmpty(undefined)).toEqual(true);
```
Hyperkid123, line 355: So the big question here is. How do we treat null and empty objects?

rvsia: It's the same as before, just copy pasted code with a new test to get it covered. :man_shrugging: :smile: If we don't want to introduce breaking changes, the question is then already answered. :smile:

PR #700, opened 2020-08-10
[manager ]feat(manager): mutate field level state on validation error

Hyperkid123:
part of #687

### DIFF
--- a/packages/form-state-manager/src/utils/use-subscription.ts
+++ b/packages/form-state-manager/src/utils/use-subscription.ts
@@ -2,7 +2,7 @@ import { useEffect, useState, useContext, useRef, useReducer } from 'react';
 import FormManagerContext from '../files/form-manager-context';
 import UseSubscription, { OnChangeEvent, SubscribtionData, Meta } from '../types/use-subscription';
 import AnyObject from '../types/any-object';
-import { fieldLevelValidator } from './validate';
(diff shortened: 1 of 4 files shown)


packages/form-state-manager/src/utils/use-subscription.ts
```diff
@@ -74,12 +74,37 @@ const useSubscription = ({ name, initialValue, clearOnUnmount, initializeOnMount
...
+        validating: false
+      }
+    }));
+  };
+
   const handleChange = (event: OnChangeEvent) => {
     const sanitizedValue = sanitizeValue(event);
     change(name, sanitizedValue);
-    // TODO Mutate field state with validation flags
+    // TODO Memoize validation results
```
rvsia, line 93: could be opt-in.. I can imagine that sometimes you want to check the same value (async validation in very dynamic applications)

Hyperkid123:
> Code looks good... but I have one request:

Could you please add a test for multiple async validations at the same time to be sure it works correctly? (so its being validated until the last validation is resolved?)

@rvsia its not ready for that. It will need a similar construct as in the manager API. I plan to do that later on after the global manager API changes as well.

packages/form-state-manager/src/utils/use-subscription.ts
Hyperkid123: Yes. But I would rather make it opt-out.

PR #703, opened 2020-08-11
feat(manager): mutate error object on field-level validation.

Hyperkid123:
part of #687

Only handles field-level validation

To enable it for form-level validation, we will need to register all validation functions and run them again after the form level validation passes. We need a way to clear form level validation errors, but do not remove field-level validation results at the same time. The only safe way I can think of is re-run all field-level validators, update the state, and possibly trigger the field render.

We will have also trigger the field render after we run form level validation for invalid fields.

### DIFF
--- a/packages/form-state-manager/src/types/validate.d.ts
+++ b/packages/form-state-manager/src/types/validate.d.ts
@@ -1,9 +1,13 @@
 import AnyObject from './any-object';
 import { ManagerApi } from './manager-api';
 
+export interface FormLevelError {
+  [key: string]: string;
+}
+
 export type Validator = (value: any, allValues: AnyObject) => undefined | string | Promise<string | undefined>;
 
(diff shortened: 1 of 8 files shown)


packages/form-state-manager/src/types/validate.d.ts
```diff
@@ -1,9 +1,13 @@
 import AnyObject from './any-object';
 import { ManagerApi } from './manager-api';
 
+export interface FormLevelError {
+  [key: string]: string;
```
rvsia, line 5: I think we should support more types. I remember I was using `<FormattedMessage />` to return translated message in Sources and it worked.

Hyperkid123: Yeah, I was thinking about this as well. I think we might have to re-visit this topic because the field level errors are locked to strings as well. And also all validator return values.

PR #704, opened 2020-08-11
fix(state): add initialValues to values

rvsia:
Flat initialValues to values

### DIFF
--- a/packages/form-state-manager/src/tests/utils/manager-api.test.js
+++ b/packages/form-state-manager/src/tests/utils/manager-api.test.js
@@ -7,8 +7,14 @@ describe('managerApi', () => {
   });
 
   it('should set initialValues', () => {
-    const managerApi = createManagerApi({ initialValues: { field: 'value' } });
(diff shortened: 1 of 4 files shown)


packages/form-state-manager/src/tests/utils/manager-api.test.js
```diff
@@ -7,8 +7,14 @@ describe('managerApi', () => {
   });
 
   it('should set initialValues', () => {
-    const managerApi = createManagerApi({ initialValues: { field: 'value' } });
-    expect(managerApi().initialValues).toEqual({ field: 'value' });
+    const initialValues = { field: 'value', nested: { some: { very: { nested: 'value', array: ['1', '2'] } } } };
```
Hyperkid123, line 10: Can you add a test case for array of objects? For a field name like these:
```
user[0].name
user[0].age
```

rvsia: Updated, you can check the new test case!

Issue #705, opened 2020-08-11
[manager] Add all necessary form options used in renderer and mappers


rvsia:
- [x]  blur: (name) => void, // calls onBlur event on field with given name
- [x]  change: (name, value) => void, // calls onChange event on field with given name
- [x]  focus: (name) => void, // calls onFocus event on field with given name

---

- [x]  getFieldState: (name) => object, // returns a state of given field, state contains input and meta information of field

@rvsia https://github.com/data-driven-forms/react-forms/pull/716

  - [x] active: false
  - [x] blur: ƒ blur()
  - [x] change: ƒ change(t)
  - [x] data: {}
  - [x] dirty: false
  - [x] dirtySinceLastSubmit: false
  - [x] error: undefined
  - [x] focus: ƒ focus()
  - [x] initial: undefined
  - [x] invalid: false

:leftwards_arrow_with_hook:  length: undefined see https://github.com/data-driven-forms/react-forms/issues/721

  - [x] modified: false
  - [x] name: "checkbox"
  - [x] pristine: true
  - [x] submitError: undefined
  - [x] submitFailed: false
  - [x] submitSucceeded: false
  - [x] submitting: false
  - [x] touched: false
  - [x] valid: true
  - [x] validating: false
  - [x] value: undefined
  - [x] visited: false

---

- [x]  getRegisteredFields: () => string[], // returns an array of field names that are rendered in DOM

---

- [x]  getState: () => object, // returns an object with whole form state. More info https://final-form.org/docs/final-form/types/FormState

https://github.com/data-driven-forms/react-forms/pull/706

  - [x] active: undefined
  - [x] dirty: false
  - [x] dirtyFields: {}
  - [x] dirtyFieldsSinceLastSubmit: {}
  - [x] dirtySinceLastSubmit: false
  - [x] error: undefined
  - [x] errors: {}
  - [x] hasSubmitErrors: false
  - [x] hasValidationErrors: false
  - [x] initialValues: {}
  - [x] invalid: false
  - [x] modified: {}
  - [x] pristine: true
  - [x] submitError: undefined
  - [x] submitErrors: undefined
  - [x] submitFailed: false
  - [x] submitSucceeded: false
  - [x] submitting: false
  - [x] touched: {}
  - [x] valid: true
  - [x] validating: false
  - [x] values: {}
  - [x] visited: {}

---

- [x]  pristine: bool, // true if the all field values is === to the initial values, false if the values are !==.
- [x]  valid: bool

[...]

PR #706, opened 2020-08-11
fix(state): match FF's getState output

rvsia:
part of https://github.com/data-driven-forms/react-forms/issues/705

### DIFF
--- a/packages/form-state-manager/src/types/any-object.d.ts
+++ b/packages/form-state-manager/src/types/any-object.d.ts
@@ -1,3 +1,7 @@
+export interface AnyBooleanObject {
+  [key: string]: boolean;
+}
+
 interface AnyObject {
   [key: string]: any;
 }
--- a/packages/form-state-manager/src/types/manager-api.d.ts
+++ b/packages/form-state-manager/src/types/manager-api.d.ts
@@ -1,4 +1,4 @@
-import AnyObject from './any-object';
+import AnyObject, { AnyBooleanObject } from './any-object';
 import { FormEvent } from 'react';
 import FieldConfig from './field-config';
 import { FormValidator } from './validate';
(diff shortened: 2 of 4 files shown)


packages/form-state-manager/src/types/manager-api.d.ts
```diff
@@ -50,8 +50,8 @@ export interface ManagerState {
   fieldListeners: AnyObject;
   active: string | undefined;
   dirty: boolean;
-  dirtyFields: Array<string>;
-  dirtyFieldsSinceLastSubmit: Array<string>;
+  dirtyFields: AnyObject;
+  dirtyFieldsSinceLastSubmit: AnyObject;
```
Hyperkid123, line 54: This should probably be a boolean object
```ts
interface Foo {
  [key: string]: boolean
}
```

packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -5,6 +5,7 @@ import CreateManagerApi, { ManagerState, ManagerApi, AsyncWatcher, AsyncWatcherR
 import AnyObject from '../types/any-object';
 import FieldConfig from '../types/field-config';
 import { formLevelValidator } from './validate';
+import { values } from 'lodash';
```
Hyperkid123, line 8: Maybe I am blind, but I can't see this used anywhere. If it is, please use direct import `lodash/values`. Or consider using `Object.values` instead.

packages/form-state-manager/src/types/any-object.d.ts
```diff
@@ -1,3 +1,7 @@
+export interface AnyBooleanObject {
+  [key: string]: any;
```
Hyperkid123, line 2: should be `:boolean` :wink:

rvsia: ..
...
....
Yep... it was made to test you. :smile:

PR #709, opened 2020-08-12
feat(manager): lift field state to global state

### LINKED ISSUE #708: [manager] Lift the field state to the manager API
In order to avoid synchronization overhead when mutating the global state or having multiple fields, we should lift the field state to the state manager. Fields will grab the state from the global manager.

Hyperkid123:
closes #708 

### Changes
- fields no longer have local state
  - the state is now stored inside the `fieldListeners` state 
- field state is updated and after the update, all fields registered under the same name are re-rendered
  - we can now use global `change` and other methods to mutate registered fields all at once
  - using the `setFieldState` will update and trigger field render
- `render` is now required registration attribute (it should have been before)

### DIFF
--- a/packages/form-state-manager/src/utils/use-subscription.ts
+++ b/packages/form-state-manager/src/utils/use-subscription.ts
@@ -1,12 +1,12 @@
-import { useEffect, useState, useContext, useRef, useReducer } from 'react';
+import { useEffect, useContext, useRef, useReducer } from 'react';
 import FormManagerContext from '../files/form-manager-context';
 import UseSubscription, { OnChangeEvent, SubscribtionData, Meta } from '../types/use-subscription';
-import AnyObject from '../types/any-object';
 import { fieldLevelValidator, isPromise } from './validate';
 import isEmpty from 'lodash/isEmpty';
 import get from 'lodash/get';
 import convertType from './convert-type';
 import { shouldExecute } from './manager-api';
+import { FieldState } from '../types/manager-api';
 
 const generateId = () => Date.now() + Math.round(Math.random() * 100000);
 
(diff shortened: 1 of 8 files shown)


packages/form-state-manager/src/utils/use-subscription.ts
```diff
@@ -157,33 +122,49 @@ const useSubscription = ({
...
     registerField({
       name,
       value: state.value,
-      getFieldState: getDetachedState,
       initializeOnMount,
       render,
       subscription,
-      internalId: state.internalId
+      internalId: state.internalId,
+      state
```
rvsia, line 167: InternalId is contained in the state so we don't need to pass it separately

packages/form-state-manager/src/utils/use-subscription.ts
```diff
@@ -157,33 +122,49 @@ const useSubscription = ({
...
+    if (firstinitialization || shouldExecute(formOptions().initializeOnMount, initializeOnMount)) {
+      value = initValue;
+    }
+
+    const state = {
+      value,
+      name,
+      meta: initialMeta(initValue),
+      internalId: generateId()
+    };
```
rvsia, line 159: I think this initialization code should be handled in `managerApi` -- schematic

i.e.:

```jsx
// manager-api

....
function registerField() {
    addToFieldListeners(internalId, render);
    if(initializeOnMount) { change(value) }; // compute initialValue/value

    if(!state[field]) {
      state[field] = createField(); // << initialMeta()
    }

   return state[field];
}
...

```

packages/form-state-manager/src/utils/use-subscription.ts
```diff
@@ -157,33 +122,49 @@ const useSubscription = ({
...
-    setState((prevState) => ({ ...prevState, value: getFieldValue(name) }));
+    setState((prevState: FieldState) => ({ ...prevState, value: getFieldValue(name) }));
   };
 
-  const valueToReturn = state.value;
+  const valueToReturn = formOptions().getFieldValue(name);
+
+  const { current: initValue } = useRef(initialValue || get(initialValues, name));
 
   useEffect(() => {
```
rvsia, line 145: We should probably move the registration code to top and do it before the rest of the code

```jsx
useState(() => {  // lazy-init... we can do it differently
    registerField(...)
})
```

This will allow us to do `const state = formOptions().getFieldState(name)` right after it and have the state synced with the global state... and we don't have to use `useRef` or store here anything anymore.

PR #711, opened 2020-08-13
feat(state): add subscribe/unsubscribe

rvsia:
parts of https://github.com/data-driven-forms/react-forms/issues/705

- also implements FormSpy

### DIFF
--- a/packages/form-state-manager/src/tests/form-spy.test.js
+++ b/packages/form-state-manager/src/tests/form-spy.test.js
@@ -1,0 +1,46 @@
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+import { mount } from 'enzyme';
+import FormManagerContext from '../files/form-manager-context';
+import createManagerApi from '../utils/manager-api';
+import FormSpy from '../files/form-spy';
+
+describe('<FormSpy />', () => {
+  it('should subscribe to changes', async () => {
+    const managerApi = createManagerApi({});
+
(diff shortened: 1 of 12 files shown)


packages/form-state-manager/src/types/form-spy.d.ts
```diff
@@ -0,0 +1,11 @@
+import { Subscription } from './use-subscription';
+import AnyObject from './any-object';
+
+type Children = (props: AnyObject) => React.ReactElement;
```
Hyperkid123, line 4: I think that we should pass specific props to the children's function. Props like value, and field metadata

Hyperkid123:
@rvsia :arrow_forward:  :basecamp:

packages/form-state-manager/src/tests/form-spy.test.js
```diff
@@ -0,0 +1,46 @@
...
+    wrapper.update();
+
+    expect(render.mock.calls.length).toEqual(2);
+
+    await act(async () => {
+      managerApi().rerender(['values']);
+    });
+    wrapper.update();
+
+    expect(render.mock.calls.length).toEqual(2);
```
Hyperkid123, line 44: You could also check that the render function receives correct values.

rvsia: the check is there on line 21. (the state is not being changed in the form manager (and we don't rerender form on chanfes) so there is no need check it multipletimes)

PR #712, opened 2020-08-13
feat(manager): create field states in manager

rvsia:
follow-up of #709 

- field state is created by the manager
- initializedFields are no longer required

### DIFF
--- a/packages/form-state-manager/src/utils/manager-api.ts
+++ b/packages/form-state-manager/src/utils/manager-api.ts
@@ -5,6 +5,8 @@ import CreateManagerApi, { ManagerState, ManagerApi, AsyncWatcher, AsyncWatcherR
 import AnyObject from '../types/any-object';
 import FieldConfig from '../types/field-config';
 import { formLevelValidator } from './validate';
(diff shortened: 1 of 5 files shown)


packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -177,26 +212,20 @@ const createManagerApi: CreateManagerApi = ({ onSubmit, clearOnUnmount, initiali
...
-      delete state.fieldListeners[field.name];
 
       if (shouldExecute(clearOnUnmount, field.clearOnUnmount)) {
         state.values[field.name] = field.value;
       }
-    } else {
-      state.fieldListeners[field.name].count -= 1;
     }
+
+    state.fieldListeners[field.name].count -= 1;
```
Hyperkid123, line 228: the `-=` and `+=` is according to some fairly buggy :man_shrugging:  (you can find a heaps of discussions around this topic). Can we switch to the long version? x = x + 1

rvsia: switched

PR #715, opened 2020-08-17
feat(manager): add async validating field flag

Hyperkid123:
part of https://github.com/data-driven-forms/react-forms/issues/687

### Changes
- moves field validation logic to manager API
- removes duplicate change state change call in the field change method
- adds async watcher logic to field level validation

### DIFF
--- a/packages/form-state-manager/src/utils/manager-api.ts
+++ b/packages/form-state-manager/src/utils/manager-api.ts
@@ -12,9 +12,9 @@ import CreateManagerApi, {
 } from '../types/manager-api';
 import AnyObject from '../types/any-object';
 import FieldConfig from '../types/field-config';
-import { formLevelValidator } from './validate';
 import { Meta } from '../types/use-subscription';
 import get from 'lodash/get';
+import { formLevelValidator, isPromise } from './validate';
 
 const isLast = (fieldListeners: AnyObject, name: string) => fieldListeners?.[name]?.count === 1;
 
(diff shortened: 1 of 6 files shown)


packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -220,6 +260,15 @@ const createManagerApi: CreateManagerApi = ({ onSubmit, clearOnUnmount, initiali
         }
       }
     };
+
+    if (state.fieldListeners[field.name]?.count === 1) {
```
rvsia, line 264: What does this condition? :thinking:

packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -212,6 +251,7 @@ const createManagerApi: CreateManagerApi = ({ onSubmit, clearOnUnmount, initiali
       ...state.fieldListeners[field.name],
       state: state.fieldListeners[field.name]?.state || createField(field, state.values[field.name]),
       count: (state.fieldListeners[field.name]?.count || 0) + 1,
+      validate: field.validate,
```
rvsia, line 254: What will happen when there are two fields with the same names and different validations?

packages/form-state-manager/src/utils/manager-api.ts
Hyperkid123: We don't want to create a new watcher instance if there are multiple fields with the same name. So its created only once.

packages/form-state-manager/src/utils/manager-api.ts
Hyperkid123: Yep, that is an issue. However, if somebody has two different validators for the same field then the validation is pointless IMO.

We could change it to be an array or something else based on the internal ID. But I think we have to think about it first. I see this similarly as an async validation concurrency. Its something that should exist, but not as a part of the library.

Hyperkid123: If we start to differentiate between field instances we would have to move the field state back from the global manager or have different states for the same field. And I think that would be a very bad decision. That basically means, having two different fields and that should be handled with different field names.

Also, we would have to change the validation behavior itself. What would be the correct global error message for that particular field if both validations would fail? Should the error message be unique or shared?

How would we change the field? We would be unable to use the global change API. I feel like if there are two fields with different validation function, then they should be different fields.

rvsia: Check the current implemenation: https://stackblitz.com/edit/snizcr?file=schema.js

All validations work at the same time.

rvsia: I think we should move the `validate` to `fields` and then just do `forEach` on them and return first error that occurs, (there still will be the shared metadata.)

PR #716, opened 2020-08-17
feat(manager): match FF's field getState

rvsia:
part of  https://github.com/data-driven-forms/react-forms/issues/705

### DIFF
--- a/packages/form-state-manager/src/utils/manager-api.ts
+++ b/packages/form-state-manager/src/utils/manager-api.ts
@@ -96,7 +96,7 @@ export function unFlatObject(obj: AnyObject): AnyObject {
 
 export const initialMeta = (initial: any): Meta => ({
   active: false,
-  data: undefined,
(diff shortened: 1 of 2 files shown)


packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -298,7 +298,13 @@ const createManagerApi: CreateManagerApi = ({ onSubmit, clearOnUnmount, initiali
 
   function getFieldState(name: string): AnyObject | undefined {
     if (state.fieldListeners[name]) {
-      return state.fieldListeners[name].state;
+      return {
+        ...state.fieldListeners[name].state,
+        ...state.fieldListeners[name].state.meta,
```
Hyperkid123, line 303: I am not sure if FF returns change blur and focus on the getFieldsState method. If it does I believe it should be prefixed by `on`.

PR #717, opened 2020-08-17
feat(manager): add reset/restart functions

rvsia:
part of https://github.com/data-driven-forms/react-forms/issues/705

todo:
- [x] should re-initilize fields

### DIFF
--- a/packages/form-state-manager/src/utils/manager-api.ts
+++ b/packages/form-state-manager/src/utils/manager-api.ts
@@ -8,7 +8,8 @@ import CreateManagerApi, {
   AsyncWatcherRecord,
   FieldState,
   Callback,
-  SubscriberConfig
+  SubscriberConfig,
+  ManagerApiFunctions
 } from '../types/manager-api';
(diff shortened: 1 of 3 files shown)


packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -212,6 +219,17 @@ const createManagerApi: CreateManagerApi = ({ onSubmit, clearOnUnmount, initiali
...
 
+  function reset(resetInitialValues?: AnyObject) {
+    state = {
+      ...state,
+      ...initialFormState(resetInitialValues || initialValues),
+      fieldListeners: state.fieldListeners,
+      registeredFields: state.registeredFields
+    };
+
+    state.registeredFields.forEach(resetFieldState);
```
Hyperkid123, line 230: Does this trigger form re-render?

rvsia: Nothing triggers rerender now. I think we should focus on rerendering/optimization after we finish with all features.

PR #719, opened 2020-08-17
Form level validation state mutation

Hyperkid123:
part of https://github.com/data-driven-forms/react-forms/issues/687

### Changes
- add state mutation after form level validation
- synchronizes form-level and field-level validation errors and state flags

### DIFF
--- a/packages/form-state-manager/src/utils/manager-api.ts
+++ b/packages/form-state-manager/src/utils/manager-api.ts
@@ -16,6 +16,7 @@ import FieldConfig from '../types/field-config';
 import { Meta } from '../types/use-subscription';
 import get from 'lodash/get';
 import { formLevelValidator, isPromise } from './validate';
+import { FormValidator, FormLevelError } from '../types/validate';
 
 const isLast = (fieldListeners: AnyObject, name: string) => fieldListeners?.[name]?.count === 1;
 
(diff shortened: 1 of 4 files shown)


packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -230,6 +232,54 @@ const createManagerApi: CreateManagerApi = ({ onSubmit, clearOnUnmount, initiali
...
+      state.invalid = true;
+    } else {
+      state.errors = {};
+      state.valid = true;
+      state.invalid = false;
+      state.error = undefined;
+      /**
+       * Fields have to be revalidated on field level to synchronize the form and field errors
+       */
+      revalidateFields(currentInvalidFields);
```
rvsia, line 273: Can we avoid this? This is not how the form currently behaves (check it [here](https://stackblitz.com/edit/jeh3gr?file=index.js)) and it brings performance issues (why to validate fields that have been already validated?)

Also, it means that active field is being validated 2x: field validation > form validation > field validation

Hyperkid123: It definitely can and should be. I wanted to postpone this after we have the memorize functionality implemented because that is a big part of it.

The validation will have have to run twice at some point (it does in the example as well) because the validation functions are different.

Once we have the memo we can "save" the active field revalidation by running the field level after the form level and check if it was validated in that cycle.

PR #731, opened 2020-08-18
fix(manager): unflat values in state

### LINKED ISSUE #728: [manager] Use nested object structure to store state values
The flat values structure causes issues with nested dynamic field registration. Currently, there will be duplicate values keys. For example, pushing to field array will result in this values object:

```json
{
  "field-array": ["foo"],
  "field-array[0]": "foo"
}
```
Only the `field-array` should exist. Nested should take values from there while subscribing as an initial value (if not set in field config)

rvsia:
closes #728

### DIFF
--- a/packages/form-state-manager/src/tests/utils/use-subscription.test.js
+++ b/packages/form-state-manager/src/tests/utils/use-subscription.test.js
@@ -117,6 +117,15 @@ describe('useSubscription', () => {
     expect(blurSpy).toHaveBeenCalledTimes(1);
   });
 
+  it('should get nested value from', () => {
+    const managerApi = createManagerApi({});
+    managerApi().change('foo[0]', 'bar');
+
+    const wrapper = mount(<DummyComponent subscriberProps={{ name: 'foo[0]' }} managerApi={managerApi} />);
+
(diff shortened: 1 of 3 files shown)


packages/form-state-manager/src/tests/utils/use-subscription.test.js
```diff
@@ -212,7 +212,7 @@ describe('useSubscription', () => {
 
       wrapper = mount(<DummyComponent subscriberProps={{ name: 'spy.nested', initializeOnMount: false }} managerApi={managerApi} />);
 
-      expect(managerApi().values).toEqual({ 'spy.nested': 'different value' });
+      expect(managerApi().values.spy.nested).toEqual('different value');
     });
   });
```
Hyperkid123, line 218: Can you add a test for this scenario:

```jsx
// not complete code just a sequence
managerApi().change('foo[1]', 'bar');
mount(<DummyComponent subscriberProps={{name: 'foo[1]'}}) ... />);
expect(w.find('input').prop('value')).toEqual('bar')
```

rvsia: added

Issue #726, opened 2020-08-18
[manager] Implement field config features


rvsia:
https://final-form.org/docs/final-form/types/FieldConfig

- [x] afterSubmit @rvsia https://github.com/data-driven-forms/react-forms/pull/759

- [x] beforeSubmit @rvsia https://github.com/data-driven-forms/react-forms/pull/759

- [x] data @rvsia https://github.com/data-driven-forms/react-forms/pull/757

- [x] defaultValue @rvsia https://github.com/data-driven-forms/react-forms/pull/750

- [x] getValidator _= validate_

- [x] initialValue

- [x] isEqual @rvsia https://github.com/data-driven-forms/react-forms/pull/736

- [x] silent @rvsia https://github.com/data-driven-forms/react-forms/pull/783

- [x] validateFields @rvsia https://github.com/data-driven-forms/react-forms/pull/736

**React variants** https://final-form.org/docs/react-final-form/types/FieldProps

- [x] afterSubmit @rvsia https://github.com/data-driven-forms/react-forms/pull/759

- [x] allowNull @rvsia https://github.com/data-driven-forms/react-forms/pull/772

- [x] beforeSubmit @rvsia https://github.com/data-driven-forms/react-forms/pull/759

- [x] children @rvsia https://github.com/data-driven-forms/react-forms/pull/776

- [x] component @rvsia https://github.com/data-driven-forms/react-forms/pull/776

- [x] data @rvsia https://github.com/data-driven-forms/react-forms/pull/757

- [x] defaultValue @rvsia https://github.com/data-driven-forms/react-forms/pull/750

- [x] format @rvsia https://github.com/data-driven-forms/react-forms/pull/752

- [x] formatOnBlur @rvsia https://github.com/data-driven-forms/react-forms/pull/752

- [x] initialValue

- [x] isEqual @rvsia https://github.com/data-driven-forms/react-forms/pull/736

- [x] multiple @rvsia https://github.com/data-driven-forms/react-forms/pull/773

- [x] name

- [x] parse @rvsia https://github.com/data-driven-forms/react-forms/pull/752

- [x] ref @rvsia https://github.com/data-driven-forms/react-forms/pull/776

- [x] render @rvsia https://github.com/data-driven-forms/react-forms/pull/776

- [x] subscription

- [x] type @rvsia https://github.com/data-driven-forms/react-forms/pull/745

- [x] validate 

[...]

PR #736, opened 2020-08-19
feat(manager): introduce isEqual and validateFields

rvsia:
part of https://github.com/data-driven-forms/react-forms/issues/726

TODO:
- [x] tests

### DIFF
--- a/packages/form-state-manager/src/utils/manager-api.ts
+++ b/packages/form-state-manager/src/utils/manager-api.ts
@@ -25,7 +25,6 @@ import FieldConfig from '../types/field-config';
 import { Meta } from '../types/use-subscription';
 import { formLevelValidator, isPromise } from './validate';
 import { FormValidator, FormLevelError } from '../types/validate';
-import { isEqual } from 'lodash';
 
 export const defaultIsEqual = (a: any, b: any) => a === b;
 
--- a/packages/form-state-manager/src/utils/manager-api.ts
+++ b/packages/form-state-manager/src/utils/manager-api.ts
@@ -25,6 +25,9 @@ import FieldConfig from '../types/field-config';
 import { Meta } from '../types/use-subscription';
 import { formLevelValidator, isPromise } from './validate';
 import { FormValidator, FormLevelError } from '../types/validate';
+import { isEqual } from 'lodash';
+
+export const defaultIsEqual = (a: any, b: any) => a === b;
 
 const isLast = (fieldListeners: AnyObject, name: string) => fieldListeners?.[name]?.count === 1;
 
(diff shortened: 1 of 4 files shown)


packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -25,6 +25,9 @@ import FieldConfig from '../types/field-config';
 import { Meta } from '../types/use-subscription';
 import { formLevelValidator, isPromise } from './validate';
 import { FormValidator, FormLevelError } from '../types/validate';
+import { isEqual } from 'lodash';
```
Hyperkid123, line 28: I don't see this used. Maybe I am just blind.

packages/form-state-manager/src/utils/manager-api.ts
```diff
@@ -425,19 +428,25 @@ const createManagerApi: CreateManagerApi = ({
...
+          dirty: !pristine
         },
         value
       }));
       state.pristine = false;
       state.dirty = true;
       validateField(name, value);
 
+      const validateFields = state.fieldListeners[name]?.validateFields;
+      validateFields && revalidateFields(validateFields);
```
Hyperkid123, line 448: The validation must be run if the `validateFields` variable is undefined. Otherwise, it will break cross-field validation since the `allValues` object was changed.

packages/form-state-manager/src/utils/manager-api.ts
rvsia: yep, the lint does not check unused vars... I could set it up.

packages/form-state-manager/src/utils/manager-api.ts
rvsia: I don't get what you mean.

It has to run when `validateFields` is defined because via this option users tell form to validate other fields.

`allValues` is referenced to `state.values` so it will be always updated.

Hyperkid123: Yes but all other fields must be revalidated on any field change by default. I am not sure if we do that at the moment. You can prove me wrong. I would add a test new for cross-field validation if possible. And we can be sure.

rvsia: No, only the changed field is validated.

`validateFields` is used to trigger validation of different field.

i.e.

```jsx
... name: 'name', validateFields: ['password'] ...
```

Password cannot contain name... so when you change `name` value, it will re-validate the `password`.

In this tests there is a case that checks only the changed field and the field in validateFields is triggered.

Hyperkid123: You are right it does not revalidate all fields. It must have changed at some point :thinking: All right then nvm this.

Hyperkid123 (approved):

Issue #721, opened 2020-08-18
[manager] Add field array support


Hyperkid123:
Implement support for dynamic array fields
- [x] introduce useFieldArray https://github.com/data-driven-forms/react-forms/pull/725
- [x] FieldArray component (wrapper around hook) https://github.com/data-driven-forms/react-forms/pull/755

Meta
- [x] length attribute https://github.com/data-driven-forms/react-forms/pull/749

Expose these methods to the form API?
- [x] push #729  
- [x] remove https://github.com/data-driven-forms/react-forms/pull/734
- [x] pop https://github.com/data-driven-forms/react-forms/pull/734
- [x] move https://github.com/data-driven-forms/react-forms/pull/741
- [x] shift https://github.com/data-driven-forms/react-forms/pull/734
- [x] swap https://github.com/data-driven-forms/react-forms/pull/741
- [x] update https://github.com/data-driven-forms/react-forms/pull/740
- [x] unshift https://github.com/data-driven-forms/react-forms/pull/743
- [x] concat https://github.com/data-driven-forms/react-forms/pull/748
- [x] insert https://github.com/data-driven-forms/react-forms/pull/747
- [x] removeBatch https://github.com/data-driven-forms/react-forms/issues/748

Public API
- [x] create common field array methods like: `const push = (name: string, value?: any) => change(name, value)` https://github.com/data-driven-forms/react-forms/issues/751
  - add the to form options to match current renderer API

rvsia:
You have missed these methods:
- [ ] concat
- [ ] insert
- [ ] removeBatch

see https://final-form.org/docs/final-form-arrays/api

rvsia:
And each field meta should get additional
- [ ] lentgh attribute

PR #743, opened 2020-08-19
feat(manager): add field array unshift method

Hyperkid123:


rvsia (approved):
great code! :traffic_light:

PR #745, opened 2020-08-19
feat(manager): add support for field.type

rvsia:
part of https://github.com/data-driven-forms/react-forms/issues/726

- checkbox and radio

### DIFF
--- a/packages/form-state-manager/src/tests/files/use-field.test.js
+++ b/packages/form-state-manager/src/tests/files/use-field.test.js
@@ -684,13 +684,35 @@ describe('useField', () => {
       });
       spy.mockReset();
 
+      // >>>>>>>>>>>>>>>> SELECT
       await act(async () => {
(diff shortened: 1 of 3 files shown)


packages/form-state-manager/src/tests/files/use-field.test.js
```diff
@@ -646,4 +646,409 @@ describe('useField', () => {
...
+        name: 'field',
+        onBlur: expect.any(Function),
+        onChange: expect.any(Function),
+        onFocus: expect.any(Function),
+        type: 'checkbox',
+        value: undefined
+      });
+    });
+
+    it('checkbox with values', async () => {
```
Hyperkid123, line 704: can you add a check that the state value has correct format? For checkbox with no value `values: {foo: true}` and for multiple choice `values: {foo: ['bar', 'baz']}`

rvsia: done

Hyperkid123 (approved):

PR #749, opened 2020-08-20
fix(manager): add length to field array meta

Hyperkid123:

### DIFF
--- a/packages/form-state-manager/src/files/use-field-array.ts
+++ b/packages/form-state-manager/src/files/use-field-array.ts
@@ -79,7 +79,7 @@ const useFieldArray: UseFieldArray = ({ name, initialValue }) => {
   };
   return {
     fields,
-    meta
+    meta: { ...meta, length: internalValue.length }
   };
 };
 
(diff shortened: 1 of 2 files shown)


packages/form-state-manager/src/files/use-field-array.ts
```diff
@@ -79,7 +79,7 @@ const useFieldArray: UseFieldArray = ({ name, initialValue }) => {
   };
   return {
     fields,
-    meta
+    meta: { ...meta, length: internalValue.length }
```
rvsia, line 82: according to docs https://final-form.org/docs/final-form/types/FieldState#length

It should be for all fields.

Hyperkid123: Well, that an `useField` attribute then. And affects non-field array fields as well.

rvsia: It should be calculated in managerApi.

Hyperkid123: Yep

Hyperkid123: We should fix it in another issue

rvsia (approved):

PR #751, opened 2020-08-20
fix(manager): create common field array api hook

Hyperkid123:

### DIFF
--- a/packages/form-state-manager/src/files/use-field-array-api.ts
+++ b/packages/form-state-manager/src/files/use-field-array-api.ts
@@ -1,0 +1,80 @@
+import { FieldArrayApi } from '../types/use-field-array-api';
+import arrayMove from '../utils/array-move';
+import { Change, GetFieldValue } from '../types/manager-api';
+
+const useFieldArrayApi = (change: Change, getFieldValue: GetFieldValue): FieldArrayApi => {
+  const enhancedGetFieldValue = (name: string): any[] => getFieldValue(name) || [];
+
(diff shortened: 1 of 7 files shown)


packages/form-state-manager/src/files/use-field-array-api.ts
```diff
@@ -0,0 +1,82 @@
+import { FieldArrayApi } from '../types/use-field-array-api';
+import { useContext } from 'react';
+import FormManagerContext from './form-manager-context';
+import arrayMove from '../utils/array-move';
+
+const useFieldArrayApi = (): FieldArrayApi => {
+  const { change, getFieldValue } = useContext(FormManagerContext);
```
rvsia, line 7: Does this really work?  `FormManagerContext.Provider` is created after this hook is used...

what about to move these functions directly to `managerApi`?

Hyperkid123: Ah, right it will cause issues in the manager. Let's just move it under the provider and pass it to the children function directly. This will always work because you can only call the hook inside the context anyway.

We could also move it to the manager API but it does not really bring anything new. It's just a bunch of wrapper functions around the manager ones. I honestly have no preference. Your call.

rvsia: Or you can take the `change` and `getFieldValue` functions from the manager and pass it as arguments to this hook. (And it actually doesn't to be a hook. :smile: )

Hyperkid123: OR that

rvsia (approved):
Thanks for the update! :sauropod:

Issue #739, opened 2020-08-19
[manager] Combine fieldConfigs


rvsia:
current behavior of FF is that it combines multiple validations to one

so validate should be stored in `fields` and it should be checked via forEach...

the same probably applies for other attributes as `isEqual` etc.

Hyperkid123:
- [ ] validate
- [ ] isEqual

rvsia:
_Note_: There should be also `afterSubmit`, `beforeSubmit` functions after/if we implement them.

Hyperkid123:
We might have to change the way of how we are triggering these functions. Specifically, the revalidate form and the fields. If we want to correctly call validators on fields with the same name but different internal ID we need to store that internal ID as well. We will have to think about it a bit more. We might have to switch away from the name identifier for certain actions and use the internal id inistead.

PR #758, opened 2020-08-20
fix(manager): allow multiple field listener validators

Hyperkid123:
part of https://github.com/data-driven-forms/react-forms/issues/739

I have added `composeValidators` but then I have realized, it's not usable for this feature. I have not added test but we will need this function later anyway.

### DIFF
--- a/packages/form-state-manager/src/files/compose-validators.ts
+++ b/packages/form-state-manager/src/files/compose-validators.ts
@@ -1,0 +1,38 @@
+/* eslint-disable no-sequences */
+import AnyObject from '../types/any-object';
+import { isPromise } from '../utils/validate';
+import ComposeValidators from '../types/compose-validators';
+
+type SyncValidator = (value: any, allValues: AnyObject) => string | undefined;
+
+/**
+ * Function that allows running mulple validation functions on a single field.
+ * Pass an array of validators as an argument.
+ * New validation will be returned, that runs all validation until first one fails, or every validator is sucessfull.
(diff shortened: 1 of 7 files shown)


rvsia (commented):
Can you please add test cases?

```jsx
registerField({name: 'field', validate: fn1})
registerField({name: 'field', validate: fn2})
```

1) two sync validations - pass & fail
2) two async validations - pass & fail
3) combine sync and async - pass & fail

Hyperkid123:
@rvsia well turns out I had to re-do the whole validation process because there was an issue with concurrency. But we should have even more powerful validation support right now. We can have unlimited number of async validators with the all-new `composeValidators` function.

rvsia:
@Hyperkid123 

> One downside is that we have run tests with setImmediate to wait for the manager api state update.

Won't be this too huge breaking change for all applications using DDF?

Hyperkid123:
@rvsia yeah it will. What we can do, is first run all the validators, check if there is an async validation and return a promise based on that.

That is an option we probably should and could do. Async validation test must have async-await/setImmediate/timeout anyway.

rvsia:
@Hyperkid123

> What we can do, is first run all the validators, check if there is an async validation and return a promise based on that.

Yeah, sounds good. Run them and just compose async validators. Do you want to do it in this PR?

Hyperkid123:
Yes I will.

Hyperkid123:
@rvsia the compose validators will only return a promise if one or more validators are asynchronous.

rvsia (approved):

PR #760, opened 2020-08-25
fix(renderer): serialize validate config to prevent infinite loop

Hyperkid123:
### Use a component like this in your form
```jsx
const TextField = ({ name, label }) => {
  const { input, meta } = useFieldApi({ name, validate: [{ type: validatorTypes.REQUIRED }] }); // the direct validate reference is the issue here
  return (
    <div>
      <label>{label}</label>
      <input {...input} />
    </div>
  );
};
```

Defining the `validate` constant directly inside the `useFieldApi` causes infinite looping. This is caused by `useEffect` which is resetting validators on `validate` change. Since in JS `[] !== []` and due to a fact that the `validate` array is a new instance in each render when used like in the example, we will be stuck in an infinite loop and crash the UI.

### Changes
- serialize validate config to prevent infinite loop

### DIFF
--- a/packages/react-form-renderer/src/files/use-field-api.js
+++ b/packages/react-form-renderer/src/files/use-field-api.js
@@ -96,7 +96,13 @@ const useFieldApi = ({ name, initializeOnMount, component, render, validate, res
         arrayValidator: calculateArrayValidator(props, validate, component, validatorMapper)
       });
     }
-  }, [validate, component, props.dataType]);
+    /**
+     * We have to stringify the validate array in order to preven infinite looping when validate was passed directly to useFieldApi
+     * const x = useFieldApu({name: 'foo', validate: [{type: 'bar'}]}) will trigger infinite looping witouth the serialize.
+     * Using stringify is acceptable here since the array is usually very small.
+     * If we notice performance hit, we can implement custom hook with a deep equal functionality.
+     */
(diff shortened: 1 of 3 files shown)


packages/react-form-renderer/src/files/use-field-api.js
```diff
@@ -96,7 +96,13 @@ const useFieldApi = ({ name, initializeOnMount, component, render, validate, res
         arrayValidator: calculateArrayValidator(props, validate, component, validatorMapper)
       });
     }
-  }, [validate, component, props.dataType]);
+    /**
+     * We have to stringify the validate array in order to preven infinite looping when validate was passed directly to useFieldApi
```
rvsia, line 100: I would convert this comment into a test case.

Hyperkid123: I am not sure how would we test that something is not causing an infinite loop.

I will add a probably useless test that the hook won't crash after we pass the validate prop directly. And maybe change it. But that probably all we can do :shrug:

rvsia: Yeah, just check it works. (Otherwise it would probably crash the whole testing because of full memory.)

rvsia (approved):

Hyperkid123:
@rvsia go away you are on vacation!

Issue #764, opened 2020-08-25
Conditionally Require Fields 


ChadLefort:
Hello, just started using this library, and I have to say it’s great and thank you for creating it! I’ve run into one small issue that I can’t figure out after looking through the docs.
​
I need to conditionally require a field input, but this field input needs to always be shown on screen. So for example, I have a select with yes/no options and another textarea field for a comment. If I select no then the comment field becomes required.
​
The only option I could sort of find was to duplicate the object and include the `validate` and `isRequired` properties on one object and exclude them on the other. Then have the conditional property toggle the visibility of fields.
​
However, this does throw a key warning. For my use case I need both name properties to be the same. The name property will be tied to an id that the backend will save the answer to.
​
Is there an option that I’m missing to conditionally include properties such as
​
```
isRequired: true,
validate: [
 {
   type: ‘required’
 }
],
```
​
on just one object? I can currently only see that I can set a value/visibility.
​
Here is a codesandbox with the schema that describes my current issue:
https://codesandbox.io/s/data-driven-forms-conditional-required-s9eme?file=/src/App.js

rvsia:
Hello, thank you for your kind words! We really appreciate it! 👍 

You can get rid of the key warning by setting a key property in the schema:

```jsx
        {
          component: componentTypes.TEXTAREA,
          name: "comment",
          label: "Comment",
          key: 'second', // some key
          isRequired: false,
          condition: {
            when: "continued-use",
            is: "yes"
          }
        }
```

And I agree, this would be the way I would implement this right now.

However, we could use this case to improve the library. There are two options what to do (and we could do both):

1) Create a new condition action

```jsx
{
   when: 'field',
   is: 'true',
   then: { setProps: { ... } },
   else: { setProps: { ... } }
}
```

2) Move `resolveProps` execution before the `useField` is used.

https://github.com/data-driven-forms/react-forms/blob/2c5978d7e62a352c2b5abb50241c291c2b0fe92b/packages/react-form-renderer/src/files/use-field-api.js#L170

this would neeed to be tranformed into

```jsx
    ...(resolveProps ? resolveProps(cleanProps, {input: { value: formOptions.values[name] ,name}, meta: formOptions.getFieldState(name)}, formOptions) : {}), // methods' names are not correct, just for the example
```

Then, it would allow to modify `validate` property from the schema.

---

Do you want to make us a PR for this? I can take a look next week or maybe @Hyperkid123 will find some time sooner.

[...]

PR #767, opened 2020-08-26
Changed validator mapper type to export object

ChadLefort:
When implementing the solution in https://github.com/data-driven-forms/react-forms/issues/764#issuecomment-680849795 I noticed that Typescript will throw an error `'validatorMapper' only refers to a type, but is being used as a value here.ts(2693).` when trying to import `validatorMapper`.

I realized that it should export a variable of this type by default since that is what is going on in https://github.com/data-driven-forms/react-forms/blob/862db36f2155cec9a6027464e9b158b4966048db/packages/react-form-renderer/src/files/validator-mapper.js#L6-L18

### DIFF
--- a/packages/react-form-renderer/src/files/form-renderer.d.ts
+++ b/packages/react-form-renderer/src/files/form-renderer.d.ts
@@ -3,7 +3,7 @@ import { FormApi, SubmissionErrors } from 'final-form';
 import { FormProps } from 'react-final-form';
 import Schema from './schema';
 import ComponentMapper from './component-mapper';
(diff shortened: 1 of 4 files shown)


rvsia:
Thank you for fixing this! 👍 

Could you please change the commit message to begin with `fix(renderer):` ? We are using a semantic release to automatically release new versions based on commit messages.

ChadLefort:
Sure thing! I should know really better since I use lerna and conventional commits on a daily basis at work haha.

rvsia (approved):

PR #775, opened 2020-09-01
fix(manager): convert initialValue/defaultValue when dataType

rvsia:
closes https://github.com/data-driven-forms/react-forms/issues/771

### DIFF
--- a/packages/form-state-manager/src/utils/convert-value.ts
+++ b/packages/form-state-manager/src/utils/convert-value.ts
@@ -1,7 +1,7 @@
 import convertType from './convert-type';
-import { DataType } from '../types/data-types';
+import { DataType, DataTypeTypes } from '../types/data-types';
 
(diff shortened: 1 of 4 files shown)


packages/form-state-manager/src/utils/convert-value.ts
```diff
@@ -0,0 +1,23 @@
+import convertType from './convert-type';
+import { DataType } from '../types/data-types';
+
+const convertValue = (value: any, dataType: DataType): any => {
```
Hyperkid123, line 4: This should only return a type specified in the `dataTypes` constant. Not any.

rvsia: With arrays and objects.. it's basically any. :D But I changed that.

Hyperkid123 (approved):

PR #777, opened 2020-09-02
feat(manager): prepare API for rendering based on state changes

rvsia:
This PR prepares API for rerendering form based on state changes

```jsx
function = () => {
   const render = prepareRerender(); // snapshots current state

   ... do some changes ...

   render(); // as the argument additional formState keys can be inserted manually, compare the old state and new state
}
```


Hyperkid123 (approved):
Hope this won't tank performance :crossed_fingers:

rvsia:
@Hyperkid123 yeah, after we implement this and test all possible cases, we can discover more performant solutions.

PR #782, opened 2020-09-03
Redirect on examples when no mapper

rvsia:
- this should prevent algolia to index MUI components twice in the search index


Hyperkid123 (approved):
:pray: 

We will see

Issue #795, opened 2020-09-09
Introduce Carbon mapper


rvsia:
**Scope**: Carbon

**Description**

Introduce Carbon mapper to DDF repo

https://react.carbondesignsystem.com/?path=/story/button--default

TODO:
- [x] basic set of components https://github.com/data-driven-forms/react-forms/pull/796
- [x] field array @rvsia https://github.com/data-driven-forms/react-forms/pull/801
- [x] duallist @rvsia  https://github.com/data-driven-forms/react-forms/pull/800
- [x] wizard @rvsia https://github.com/data-driven-forms/react-forms/pull/803
- [x] validateOnMount @rvsia https://github.com/data-driven-forms/react-forms/pull/812
- [x] explore comboBox instead of select (supports filtering, isClearable)
    - observation: **ComboBox** seems more like a better text input so I won't looking into it further.
- [x] style form template according to docs (48px margin for title/button, gaps, etc.) @rvsia https://github.com/data-driven-forms/react-forms/pull/813
- [x] timepicker does not work at all (timezones, format selector) @rvsia  https://github.com/data-driven-forms/react-forms/pull/814
- [x] create helperText wrapper for components not supporting it @rvsia https://github.com/data-driven-forms/react-forms/pull/815
- [x] treeshaking https://github.com/data-driven-forms/react-forms/pull/796
- [x] documentation
   - [x] mapper page https://github.com/data-driven-forms/react-forms/pull/796
   - [x] component examples https://github.com/data-driven-forms/react-forms/pull/796
   - [x] specific texts for components @rvsia https://github.com/data-driven-forms/react-forms/pull/808

rvsia:
# Changelog of changes PF3 > Carbon `WORK IN PROGRESS`

## No `isRequired`

Carbon recommends to mark only optional fields (see [here](https://www.carbondesignsystem.com/patterns/forms-pattern#building-a-form), so there is no `isRequired` prop.)

**Required field**

:stop_sign: 

```jsx
{
   component: 'text-field',
   label: 'login',
   isRequired: true,
}
```

:ok: 

```jsx
{
   component: 'text-field',
   label: 'login',
}
```

**Optional field**

:stop_sign: 

```jsx
{
   component: 'text-field',
   label: 'description',
}
```

:ok: 

```jsx
{
   component: 'text-field',
   label: 'description (optional)',
}
```

The Carbon documentation also suggests **to group** optional fields to avoid repeatable `(optional)` text.

## Select

[...]

PR #796, opened 2020-09-09
feat(carbon): introduce carbon mapper

rvsia:
TODO:
- :stop_sign:  helperText - switch, checkbox, etc.
    - would be handled seperately

---
- [x] select component
- [x] tests

part of #795


packages/carbon-component-mapper/src/files/date-picker.d.ts
```diff
@@ -0,0 +1,13 @@
+import { FormGroupProps } from "./form-group";
+import { UseFieldApiComponentConfig, AnyObject } from "@data-driven-forms/react-form-renderer";
+
+interface InternalDatePickerProps extends AnyObject, React.HTMLProps<HTMLInputElement> {
```
Hyperkid123, line 4: This does not extend carbon date picker props.

packages/carbon-component-mapper/src/files/plain-text.d.ts
```diff
@@ -0,0 +1,10 @@
+import { ReactNode } from "react";
+
+export interface PlainTextProps {
+  label: ReactNode;
+  element?: string;
```
Hyperkid123, line 5: There should be a list somewhere of acceptable element strings.

packages/carbon-component-mapper/src/files/radio.d.ts
```diff
@@ -0,0 +1,21 @@
+import { FormGroupProps } from "./form-group";
+import { UseFieldApiComponentConfig, AnyObject } from "@data-driven-forms/react-form-renderer";
+import { ReactNode } from "react";
+
+export interface RadioOption extends AnyObject {
+    value: any;
+    label?: ReactNode;
+}
+
+interface InternalRadioProps extends AnyObject {
```
Hyperkid123, line 10: This should extend carbon radio props.

packages/carbon-component-mapper/src/files/radio.d.ts
```diff
@@ -0,0 +1,21 @@
...
+    isDisabled?: boolean;
+    label?: ReactNode;
+    options: RadioOption[];
+    FormGroupProps?: AnyObject;
```
Hyperkid123, line 14: This should be the form group interface declared above.

packages/carbon-component-mapper/src/files/select.d.ts
```diff
@@ -0,0 +1,19 @@
+import { FormGroupProps } from "./form-group";
+import { UseFieldApiComponentConfig, AnyObject } from "@data-driven-forms/react-form-renderer";
+import { ReactNode } from "react";
+
+export interface SelectOption extends AnyObject {
+  value?: any;
+  label: ReactNode;
+}
+
+interface InternalSelectProps extends React.HTMLProps<HTMLSelectElement> {
```
Hyperkid123, line 10: Again this should extend the carbon select props

Hyperkid123 (commented):
A lot of the TS module definitions are not extending the carbon component interface. That should be fixed.

packages/carbon-component-mapper/src/files/date-picker.d.ts
rvsia: @Hyperkid123 Carbon is not using typescript to define props.

packages/carbon-component-mapper/src/files/plain-text.d.ts
rvsia: Do you think it’s possible to generate this? Otherwise, it can be really anything, because Carbon is using global styles and does not have specific components for texts.

packages/carbon-component-mapper/src/files/radio.d.ts
rvsia: This is a little bit confusing, but FormGroup interface should be probably named “commonProps” because in this mapper there is no FormGroup component. (The name is generated via the generate command.) This prop is for Carbon’s FormGroup component.

packages/carbon-component-mapper/src/files/date-picker.d.ts
Hyperkid123: :facepalm:

packages/carbon-component-mapper/src/files/plain-text.d.ts
Hyperkid123: Well, the plain text should be anything that uses text. So we are talking about headings, paragraphs, labels, etc. I think either PF4 or PF3 mappers have the list so we can make it a common props. Or maybe it is already in common props. By leaving this as a string, we are basically allowing it to be any element and we are getting to a `dangerouslySetInnerHTML` territory.

Hyperkid123:
@rvsia 
> A lot of the TS module definitions are not extending the carbon component interface. That should be fixed.

Alright, I did not expect this. If carbon is not using TS definitions, we can't provide them. Therefore I would not generate TS typings at all for this module. We can't use HTML element props since the APIs do not match and using just `AnyObject` is pointless and I don't think we want to declare all the modules on our own.

Hyperkid123:
@rvsia we have a bit of luck.

https://github.com/DefinitelyTyped/DefinitelyTyped#readme does include types for https://www.npmjs.com/package/@types/carbon-components-react. Can you check if this matches the carbon API? we could use that instead.

packages/carbon-component-mapper/src/files/plain-text.d.ts
rvsia: I don't think it matters here. Programmers with the access to typescript can actually do everything they want to do, so I don't see why to limit them. 

And to prevent outside attacks we should do a real-time check in the render of the function, but we are not doing that in other mappers and maybe we should let application teams to handle that. (With a component wrapper or with a `resolveProps` function.)

rvsia:
@Hyperkid123 used the definitely typed package for extending the carbon props.

packages/carbon-component-mapper/src/files/text-field.d.ts
```diff
@@ -0,0 +1,13 @@
+import { FormGroupProps } from "./form-group";
+import { UseFieldApiComponentConfig } from "@data-driven-forms/react-form-renderer";
+
+import { TextInputProps } from 'carbon-components-react';
+
+interface InternalTextFieldProps extends TextInputProps {
```
Hyperkid123, line 6: This is not necessary you can directly reference carbon props in the `TextFieldProps` type.

packages/carbon-component-mapper/src/files/textarea.d.ts
```diff
@@ -0,0 +1,13 @@
+import { FormGroupProps } from "./form-group";
+import { UseFieldApiComponentConfig } from "@data-driven-forms/react-form-renderer";
+
+import { TextAreaProps } from 'carbon-components-react';
+
+interface InternalTextareaProps extends TextAreaProps {
```
Hyperkid123, line 6: ^^

Hyperkid123 (approved):

PR #800, opened 2020-09-14
feat(carbon): add dual list component

rvsia:
part of https://github.com/data-driven-forms/react-forms/issues/795

TODO:
- [x] tests
- [x] treeshake carbon react icons

carbonduallist

### DIFF
--- a/packages/carbon-component-mapper/demo/index.js
+++ b/packages/carbon-component-mapper/demo/index.js
@@ -45,7 +45,38 @@ class App extends React.Component {
             componentMapper={componentMapper}
             FormTemplate={(props) => <FormTemplate {...props} showFormControls={this.state.additionalOptions.showFormControls} />}
             onCancel={console.log}
-            schema={this.state.schema}
+            //schema={this.state.schema}
+            schema={{
+              fields: [
+                {
+                  component: 'dual-list-select',
+                  name: 'dual-list',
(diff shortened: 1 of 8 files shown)


Hyperkid123:
@rvsia I think we need to clean this up a bit.

- The dimensions of the list change on the hover event (the item gets taller). Can we do something about that?
- We need some spacing between the move buttons
- The move all buttons should be next to each other -> moveRight, moveAllRight, moveAllLeft, moveLeft 
- The dropdown icon looks out of place. Probably should be aligned to the center of the filter input and have more spacing around it. And it does not look like its interactable.
- The middle column should have a bit more spacing from the sides.

rvsia:
Ad 1) it's controlled from Carbon itself.
Ad 2) Ok
Ad 3) Ok
Ad 4) Dropdown icon cannot be aligned, the offset is set in the svg
Ad 5) Ok
Ad 6) Carbon

updated

Hyperkid123:
@rvsia could you also provide a picture of this on smaller screen?

rvsia:
image

Hyperkid123:
@rvsia ok the only mini nitpick I have is that the arrow icons on buttons should also change direction towards the list.

rvsia:
done

Hyperkid123 (approved):

PR #807, opened 2020-09-16
fix(renderer): export parse condition function

### LINKED ISSUE #802: Making condition evaluation more generic
I'm struggling to reimplement the conditional validators described [here](https://github.com/data-driven-forms/react-forms/issues/149#issuecomment-539401224) using the validator mapper in my own codebase. It would be nice to have an `evaluateCondition` that would just give me true or false without passing any context inside:

```js
import { validatorMapper, evaluateCondition } from '@@ddf';

const decorate = fn => ({ condition, ...rest }) => {
  if (!evaluateCondition) return true;
  return fn(rest);
}

const myValidatorMapper = Object.keys(validatorMapper).reduce((obj, key) => ({
   ...obj,
   [key]: decorate(validatorMapper[key]),
}) ,{});

```

Hyperkid123:
closes #802

### DIFF
--- a/packages/react-form-renderer/src/files/parse-condition.d.ts
+++ b/packages/react-form-renderer/src/files/parse-condition.d.ts
@@ -1,0 +1,6 @@
+import { AnyObject } from "./common";
+import { ConditionDefinition } from "./condition";
+
+export type ParseCondition = (condition: ConditionDefinition, values: AnyObject) => void;
+declare const parseCondition: ParseCondition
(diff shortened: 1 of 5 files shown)


rvsia (approved):
Not sure how this is going to enhance custom validators, but sure, there is no reason to keep it internal.

However, we should not encourage people to use it without a proper testing, as it's an internal function, so it can be changed anytime.

PR #815, opened 2020-09-24
Carbon helper text

rvsia:
- adds helperText and errorTexts where are not natively supported

### DIFF
--- a/packages/carbon-component-mapper/src/files/checkbox.js
+++ b/packages/carbon-component-mapper/src/files/checkbox.js
@@ -7,21 +7,35 @@ import { Checkbox as CarbonCheckbox, FormGroup } from 'carbon-components-react';
 
 import WithDescription from '../common/with-description';
 import prepareProps from '../common/prepare-props';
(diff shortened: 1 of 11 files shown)


packages/carbon-component-mapper/src/files/checkbox.js
```diff
@@ -7,21 +7,35 @@ import { Checkbox as CarbonCheckbox, FormGroup } from 'carbon-components-react';
...
+    <HelperTextBlock helperText={helperText} errorText={showError && error} />
+  </FormGroup>
 );
 
 Wrapper.propTypes = {
   label: PropTypes.node,
   children: PropTypes.node,
-  description: PropTypes.node
+  description: PropTypes.node,
+  helperText: PropTypes.node,
```
Hyperkid123, line 23: Are these new props part of the component typings?

rvsia: Yeah, they are in the formgroup definition.

Hyperkid123 (approved):

PR #821, opened 2020-10-05
feat(renderer): dynamic when attribute

### LINKED ISSUE #820: Conditionals in Field Arrays
**Scope**: Schema / FormRenderer

**Description**: Would love the ability to put up conditionals within field arrays. Currently now the condition target is an exact path, working with field arrays the field will be an array [0], [1] etc, thus making it impossible to dynamically alter the condition based on every new "row" of field array directly in the schema.

**Schema**
Short example code (disregard obvious type errors)
```js
const fieldOptions = [...fieldTypeOptions];
const fieldSchema = [...schemaFields];
const searchField = {
    component: "field-array",
    name: "FieldArr",
    fields: [
        {
            FormFieldGridProps: { md: 12 },
            component: "autocomplete-async",
            label: "Field type",
            name: "FieldType",
            options: fieldOptions,
        },
        ...Object.entries(fieldSchema).map((entry) => {
            const [key, field] = entry;
            const target = fieldOptions.find((item) => item.value == key);

            return Object.assign(field, {
                FormFieldGridProps: { md: 12 },
                condition: {
                    when: "FieldType",
                    is: target ,
                    then: { visible: true },
                },
            });
        }),
    ],
};
```

The condition will not match since FieldType in the instance will be "FieldArr[0].FieldType" and continuing on the next added field array the target will be "FieldArr[0].FieldType", making it impossible to dynamically match field array conditions based on the current instance "row"

Would be great if you could target the path based on the current index of field array, for example:
```js
condition: {
    when: (fieldIndex, rowData) => `FieldArr[${fieldIndex}].FieldType[0]`,
    is: target,
    then: { visible: true },
},
```

rvsia:
closes #820

- when in condition can be a function that gets the current field as the first argument

**Function**

*currentField => string | string[]*

A function receiving the current field as the first argument. Can be used to get dynamically created field name:

```jsx
{
  when: (field) => `${getFieldNumber(field)}.name`
}
/*
  field = {
    name: 'person[14]',
  }
*/
```

A function in an array cannot return another array.

### DIFF
--- a/packages/react-form-renderer/src/files/condition.d.ts
+++ b/packages/react-form-renderer/src/files/condition.d.ts
@@ -1,10 +1,15 @@
+import Field from "./field";
+
 export interface ActionResolution {
   visible?: boolean;
   set?: object; // TO DO specify this
 }
 
+export type InnerWhenFunction = (currentField: string) => string;
+export type WhenFunction = (currentField: string) => string | string[] | InnerWhenFunction[];
+
 export interface ConditionProp {
-  when?: string | string[];
+  when?: string | string[] | WhenFunction | InnerWhenFunction[];
   is?: any;
   isNotEmpty?: boolean;
   isEmpty?: boolean;
@@ -28,6 +33,7 @@ export interface ConditionProps {
   values: object;
   children: React.ReactChildren;
   condition?: ConditionDefinition | ConditionDefinition[];
+  field: Field;
 }
 
 declare const Condition: React.ComponentType<ConditionProps>;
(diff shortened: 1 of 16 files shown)


packages/react-form-renderer/src/files/condition.d.ts
```diff
@@ -28,6 +33,7 @@ export interface ConditionProps {
   values: object;
   children: React.ReactChildren;
   condition?: ConditionDefinition | ConditionDefinition[];
+  field: Field;
```
Hyperkid123, line 36: Why was this added? I see that you have swapped it with `formOptions`. Did we forget to include for options inside this typing?

rvsia: `formOptions` were not used at all. the `field` is a object passed to `when` function.

Hyperkid123: :+1:

Hyperkid123 (approved):

PR #822, opened 2020-10-06
feat(manager): implement async onSubmit

rvsia:
allows to use async function for submits

### DIFF
--- a/packages/form-state-manager/src/types/manager-api.d.ts
+++ b/packages/form-state-manager/src/types/manager-api.d.ts
@@ -24,7 +24,7 @@ export type HandleSubmit = (event?: FormEvent) => void;
 export type RegisterField = (field: FieldConfig) => void;
 export type UnregisterField = (field: Omit<FieldConfig, 'render'>) => void;
 export type GetState = () => ManagerState;
-export type OnSubmit = (values: AnyObject) => void;
+export type OnSubmit = (values: AnyObject) => any;
 export type GetFieldValue = (name: string) => any;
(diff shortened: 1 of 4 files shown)


packages/form-state-manager/src/types/manager-api.d.ts
```diff
@@ -24,7 +24,7 @@ export type HandleSubmit = (event?: FormEvent) => void;
 export type RegisterField = (field: FieldConfig) => void;
 export type UnregisterField = (field: Omit<FieldConfig, 'render'>) => void;
 export type GetState = () => ManagerState;
-export type OnSubmit = (values: AnyObject) => void;
+export type OnSubmit = (values: AnyObject) => any;
```
Hyperkid123, line 27: How do we want to handle submission errors? I think we want some global error object which we can the render on the top of the form template for example, or we should expect an error object with the same structure as current validation errors have,

rvsia: FThere is `submitError` attribute on the state object that is used to store form level error. 

https://final-form.org/docs/final-form/types/FormState#submiterror

And I just see there is a `FORM_ERROR` key to store the error (right now I store there the return of submit function). Should we follow this?

rvsia: How FF does it:

everything returned go to `submitErrors`
attribute with key `FORM_ERROR` goes to `submitError`

I will change it here.

Hyperkid123 (approved):

PR #829, opened 2020-10-12
fix(pf3): allow select null value

### LINKED ISSUE #615: Option with empty value never gets selected in the DDF select
**Scope**: DDFv1 PF3

**Description**
> In select fields, when there's a nulll/0/empty value for an option, it never gets preselected if the field matches the value and the placeholder is shown. If you try to select it manually, the same thing happens, the default empty value tied to the placeholder takes preference.

https://github.com/ManageIQ/manageiq-ui-classic/issues/7174

Hyperkid123:
closes #615

If null option was selected, the label was not showing as selected value (although the null value was submitted). Snapshots are updated due to the dependency version update.

falsey-select

@skateman the other values (0, -1, ...) were working without any modification.

### DIFF
--- a/packages/pf3-component-mapper/demo/index.js
+++ b/packages/pf3-component-mapper/demo/index.js
@@ -24,6 +24,23 @@ const loadOptions = () => new Promise((res) => {
 const selectSchema = {
   fields: [{
     component: 'select',
+    name: 'falsey-values',
+    label: 'Falsey values',
+    options: [{
+      label: 'Zero',
+      value: 0
+    }, {
+      label: 'Minus one',
+      value: -1
+    }, {
+      label: 'Empty',
+      value: 'empty'
+    }, {
+      label: 'Null',
+      value: null
+    }]
+  }, {
+    component: 'select',
     name: 'async-single',
     label: 'Async single',
     isMulti: true,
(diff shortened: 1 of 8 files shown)


packages/pf3-component-mapper/demo/index.js
```diff
@@ -121,7 +138,7 @@ const App = () => {
           onSubmit={console.log}
           componentMapper={componentMapper}
           FormTemplate={(props) => <FormTemplate {...props} />}
-          schema={schema}
+          schema={{fields: [selectSchema.fields[0]]}}
```
rvsia, line 141: I would keep demo unchanged to allow using buttons.

packages/pf3-component-mapper/src/files/select.js
```diff
@@ -36,9 +36,14 @@ Select.propTypes = {
   description: PropTypes.node,
   placeholder: PropTypes.string,
   isDisabled: PropTypes.bool,
-  inputAddon: PropTypes.shape({ fields: PropTypes.array })
+  inputAddon: PropTypes.shape({ fields: PropTypes.array }),
+  allowNull: PropTypes.bool,
```
rvsia, line 40: does this do something?

Hyperkid123: Yes. The final form field treats null as falsey value and its converted to `undefined` by default.

rvsia (approved):

skateman (approved):
The Seal of Approval

PR #836, opened 2020-10-15
 feat(renderer): add useWarnings functionality

### LINKED ISSUE #835: [CustomValidator] Custom warning instead of custom error
**Scope**: CustomValidator functionality

**Description**

I'm working with forms that require showing "alerts/warnings" on some fields after validating them.
For example, after entering a text in a field, the validation of that field should be triggered and if the value of the field matches a previously fetched array of values, the field should show a warning message of some sort (eg. "the value already exists").

The thing is that the message shouldn't be treated as a validation error, meaning that the form should still be submittable (if there are no other fields with an actual error).

If I have understood the docs correctly, there is no way for achieving this currently, aka, Validators can trigger "error" or "not error", but they can't trigger "kind of an error, but not error". Is that correct?

If that is not correct, how can I handle a case like the one I just described?

**Schema**

N/A

rvsia:
closes #835 

I have tried several ways how to store warnings and not to use state (`useRef`, `meta.data` via mutator), however I have encounter several issues with the initial validation, where both solutions didn't triggered a re-rerender and who knows how it would handle async functions.

So I implemented it using `useState` and to prevent performance issues, I created a `useWarnings` prop, that turn this functionality on. However, this prop **cannot be changed at runtime** because of hooks ordering. That's one possible issue.

# Validator warnings

If you don't want to block the form from submitting when a validator fails, you can convert the validator into a warning validator.

## 1. Enable warnings

In the field, set `useWarnings` to `true`. This feature can impact performace, so it was made to be opt-in. Please, do not change this value run-time as it could introduce bugs because of React hooks ordering.

```jsx
{
    component: 'text-field',
    useWarnings: true,
    ...
}
```

### Object validator

When you want to return an object validator as a function, just add `warning: true` to its configuration.

```jsx
{
    ...,
    validate: [{ type: 'required', warning: true }]
}
```

### Function validator

When using a function as the validator, the function has to return an object with the following format:

```
{
    warning: true,
    error: 'some message'
}
```

## 3. Get warning

A warning is stored in `meta.warning`. This value is specific for each field and fields with the same name do not share this. Also, this value is not reachable via `resolveProps` as it's computed just after the `resolveProps` function.

```jsx
const TextFieldWarning = (props) => {
  const { input, meta, ...rest } = useFieldApi(props);
  return (
    <div>
      <input {...input} {...rest} />
      {meta.warning && <div id="warning">{meta.warning}</div>}
    </div>
  );
};
```

### Convert to error

By default, Data Driven Forms mappers are not supporting showing warnings. However you can convert the warning into an error - set `convertWarningToError` to `true`. This will replace `meta.error` with `meta.warning`, but `valid` etc. will be unchanged.

```jsx
{
    name: 'warning-as-error',
    component: 'text-field',
    useWarnings: true,
    convertWarningToError: true,
    validate: [{type: 'required', warning: true}]
}
```

TODO: 
- [x] variant for type validators
- [x] tests
- [x] docs

### DIFF
--- a/packages/react-form-renderer/src/files/use-field-api.js
+++ b/packages/react-form-renderer/src/files/use-field-api.js
@@ -21,16 +21,34 @@ const calculateArrayValidator = (props, validate, component, validatorMapper) =>
   }
 };
 
-const calculateValidate = (props, validate, component, validatorMapper) => {
+const calculateValidate = (props, validate, component, validatorMapper, setWarning, useWarnings) => {
   if ((validate || props.dataType) && componentTypes.FIELD_ARRAY !== component) {
-    return composeValidators(getValidate(validate, props.dataType, validatorMapper));
+    const validateFn = composeValidators(getValidate(validate, props.dataType, validatorMapper));
+
+    if (useWarnings) {
+      return async (...args) => {
+        setWarning(undefined);
+
+        const result = await validateFn(...args);
+
+        if (result?.type === 'warning') {
+          setWarning(result.error);
+
+          return;
+        }
+
+        return result;
+      };
+    }
+
+    return validateFn;
   }
 };
 
-const init = ({ props, validate, component, validatorMapper }) => ({
+const init = ({ props, validate, component, validatorMapper, setWarning, useWarnings }) => ({
   initialValue: calculateInitialValue(props),
   arrayValidator: calculateArrayValidator(props, validate, component, validatorMapper),
-  validate: calculateValidate(props, validate, component, validatorMapper),
+  validate: calculateValidate(props, validate, component, validatorMapper, setWarning, useWarnings),
   type: assignSpecialType(component)
 });
 
@@ -68,9 +86,7 @@ const createFieldProps = (name, formOptions) => {
 
 const useFieldApi = ({ name, initializeOnMount, component, render, validate, resolveProps, useWarnings, convertWarningToError, ...props }) => {
   const { validatorMapper, formOptions } = useContext(RendererContext);
-
-  // eslint-disable-next-line react-hooks/rules-of-hooks
-  const [warning, setWarning] = useWarnings ? useState() : [undefined, () => undefined];
+  const [warning, setWarning] = useState();
 
   const { validate: resolvePropsValidate, ...resolvedProps } = resolveProps
     ? resolveProps(props, createFieldProps(name, formOptions), formOptions) || {}
@@ -80,7 +96,7 @@ const useFieldApi = ({ name, initializeOnMount, component, render, validate, res
 
   const [{ type, initialValue, validate: stateValidate, arrayValidator }, dispatch] = useReducer(
     reducer,
-    { props: { ...props, ...resolvedProps }, validate: finalValidate, component, validatorMapper },
+    { props: { ...props, ...resolvedProps }, validate: finalValidate, component, validatorMapper, setWarning, useWarnings },
     init
   );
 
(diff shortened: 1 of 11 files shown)


packages/react-renderer-demo/src/pages/schema/validator-warnings.md
```diff
@@ -0,0 +1,75 @@
...
+      <input {...input} {...rest} />
+      {meta.warning && <div id="warning">{meta.warning}</div>}
+    </div>
+  );
+};
+```
+
+### Convert to error
+
+By default, Data Driven Forms mappers are not supporting showing warnings. However you can convert the warning into an error - set `convertWarningToError` to `true`. This will replace `meta.error` with `meta.warning`, but `valid` etc. will be unchanged.
```
Hyperkid123, line 63: I am not sure if this is a good approach. I think we can just dump the waring into a helper text. And have a priority 
```
error > warning
```
This would require some change to the compose validator and storing the warning to wait for the full result.

 I don't see a reason why would anybody start using soft validation and don't want to display it in the UI.

I think we should keep it as simple as possible. `convertWarningToError` is maybe just too much as well. As I understand it correctly, it only changes the "visual" representation of the field but the validation still fails. We have global props for that.

packages/react-form-renderer/src/files/use-field-api.js
```diff
@@ -66,9 +66,12 @@ const createFieldProps = (name, formOptions) => {
   };
 };
 
-const useFieldApi = ({ name, initializeOnMount, component, render, validate, resolveProps, ...props }) => {
+const useFieldApi = ({ name, initializeOnMount, component, render, validate, resolveProps, useWarnings, convertWarningToError, ...props }) => {
   const { validatorMapper, formOptions } = useContext(RendererContext);
 
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const [warning, setWarning] = useWarnings ? useState() : [undefined, () => undefined];
```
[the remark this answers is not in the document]
rvsia: So do you want to introduce `useState` that 99% of the components will never use?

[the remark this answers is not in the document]
Hyperkid123: Better than potential bugs.

packages/react-renderer-demo/src/pages/schema/validator-warnings.md
rvsia: `I think we can just dump the waring into a helper text.` I don't think that's a good idea. `helperText` has totally different meaning and you should be able to differentiate between helperText and warnings. helperText is also not-common props and you can have components without it.

`I don't see a reason why would anybody start using soft validation and don't want to display it in the UI.` It doesn't mean that users do not want to use soft validation without UI, it means that you can have different behavior for error / warning. This props was made to be able to use warnings in old components, that don't know what a warning is.

packages/react-form-renderer/src/files/use-field-api.js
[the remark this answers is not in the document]
rvsia: I think no bugs can occur if users won't change this prop runtime. But ok.

packages/react-renderer-demo/src/pages/schema/validator-warnings.md
Hyperkid123: > helperText has totally different meaning and you should be able to differentiate between helperText and warnings

This is actually not entirely true, if you look at current mappers they all use helper text to display errors. Pf4 actually even has a warning state and it consumes the helper text. We can add a simple CSS for MUI (obviously customizable) which will indicate the warning state. I think you are thinking of a description. But helper text should describe the current state of the input. So if there is a warning/error/success messages, that should be displayed inside helper text.

The only additional attribute for this feature should be the flag inside the validator or the result that's it.

> This props was made to be able to use warnings in old components, that don't know what a warning is.

Well yes, but it does not really work. The component will be reporting a warning state but actually, the form is in an error state. So it will give users false feedback. The fact that mappers don't know how to react to the state is a different issue and well, not all mappers will use this state. That is not an issue. The renderer should not try to migrate mappers for new features. We will have to adjust mappers so they can interpret the feature. I think we can simply start with MUI mapper. And PF4 because that actually has a warning state.

rvsia: Ok, I will remove this option at all. You are right, this should be handled by mappers separately, I just wanted to introduce a simple bridge until mapper will be ready.

We should really progress with the state manager, where we can implement warning right into the form state and all these issues will disappear. :snail: :boom:

Hyperkid123: Yeah, I would love to get back to the state manager... Who knows maybe I'll have more time work on that now that we have the lockdown again :smile:

Hyperkid123 (approved):

PR #845, opened 2020-10-21
feat(manager): introduce warnings feature

### LINKED ISSUE #843: [manager] Implement warning
**Scope**: form state manager

**Description**

Implement warning functionality, so warnings are stored in the form state and no additional useState is required

rvsia:
fixes #843 

- it follows the same format as the renderer, however it's no longer opt-in (so we could get rid of `useWarnings` in the next version)

### DIFF
--- a/packages/form-state-manager/src/files/compose-validators.ts
+++ b/packages/form-state-manager/src/files/compose-validators.ts
@@ -10,15 +10,15 @@ import ComposeValidators from '../types/compose-validators';
  * @returns {Function} New validation function
  */
 const composeValidators: ComposeValidators = (validators = []) => (value, allValues) => {
-  const promises: Promise<string | undefined>[] = [];
+  const promises: Promise<any>[] = [];
   let index = 0;
-  let error: string | undefined;
+  let error: any;
   while (validators.length > 0 && !error && index < validators.length) {
     const result = validators[index](value, allValues);
     if (isPromise(result)) {
-      promises.push(result as Promise<string | undefined>);
+      promises.push(result as Promise<any>);
     } else {
-      error = result as string | undefined;
+      error = result as any;
     }
 
     index = index + 1;
--- a/packages/form-state-manager/src/types/compose-validators.d.ts
+++ b/packages/form-state-manager/src/types/compose-validators.d.ts
@@ -1,6 +1,13 @@
 import { Validator } from './validate';
 import AnyObject from './any-object';
 
-export type ComposeValidators = (validators: Validator[]) => (value: any, allValues: AnyObject) => Promise<any> | any;
+export interface WarningObject<T = string | undefined> {
+  type: string;
+  error: T;
+}
+
+export type ComposeValidators<T = WarningObject | string | undefined> = (
+  validators: Validator[]
+) => (value: any, allValues: AnyObject) => Promise<T> | T;
 
 export default ComposeValidators;
--- a/packages/form-state-manager/src/types/compose-validators.d.ts
+++ b/packages/form-state-manager/src/types/compose-validators.d.ts
@@ -2,7 +2,7 @@ import { Validator } from './validate';
 import AnyObject from './any-object';
 
 export interface WarningObject<T = string | undefined> {
-  type: string;
+  type: 'warning';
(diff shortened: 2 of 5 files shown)


packages/form-state-manager/src/files/compose-validators.ts
```diff
@@ -10,15 +10,15 @@ import ComposeValidators from '../types/compose-validators';
  * @returns {Function} New validation function
  */
 const composeValidators: ComposeValidators = (validators = []) => (value, allValues) => {
-  const promises: Promise<string | undefined>[] = [];
+  const promises: Promise<any>[] = [];
```
Hyperkid123, line 13: This should not be any. It should be `string | undefined | WarningResult` or something.

rvsia: I think it can - what about react nodes for example?

Hyperkid123: We must have a strict declaration of the error message format. Anybody can make nodes from it in mapper. We should not expect that API will return react node...

Hyperkid123: This is a function that is used by the state manager and not the react binding for it. Imagine that the manager is not used withing react context. How would they consume nodes?

rvsia: I am just saying that error can be now anything and it works - see the example below:

see https://stackblitz.com/edit/5q2uq8?file=index.js

So I don't see any reason why to limit it when there is no computation/manipulation (so it cannot be break if you put there anything) with this value.

Hyperkid123: Ofc it can be anything. I will rephrase the question. What is in our docs? Validators should return `undefined` or `string` or a warning object. Why do we expect anything? Do we remove that rule?

Hyperkid123: If you want to expand on our implementation, make the validator a generic type and make it extend the return type with a user-defined type. That way we can keep it strictly and safely defined.

Hyperkid123: Or at minimum make it `string | undefined | WarningResult | any`.

rvsia: https://data-driven-forms.org/schema/custom-validator

There is already `any`.

Real example: In sources,  `<FormattedMessage />` is put there.

Hyperkid123: Then it is wrong. And it should be fixed. `WarningResult` is an important structure because it cannot be used as a node and will crash in runtime.

rvsia: I don't understand. What will crash in runtime?

Hyperkid123: I meant compile-time sorry. The compose validator is an exported function. Yes, we do catch the warning object internally and assing the warning to the meta-object, but if used outside it can return an object that is not usable for render. When using TS, you will get an error that the return type is any and the code will not compile.

By changing the function type to a generic function, you will avoid additional typecasting and let the user define return type of their function. Therefore no compile time errors and typecasting:
```ts
type defaultType = string | warning | any
function ComposeValidators<T = defaultType>(arg: foo): T {
  return arg;
}
```

packages/form-state-manager/src/types/compose-validators.d.ts
```diff
@@ -1,6 +1,13 @@
 import { Validator } from './validate';
 import AnyObject from './any-object';
 
-export type ComposeValidators = (validators: Validator[]) => (value: any, allValues: AnyObject) => Promise<any> | any;
+export interface WarningObject<T = string | undefined> {
+  type: string;
```
Hyperkid123, line 5: I think this should a list of currently supported states. Right now it's just a "warning"

rvsia: Yeah.. right now it's just one exact string :smile:

Hyperkid123 (approved):

PR #846, opened 2020-10-21
feat(carbon): implemented optional isRequired for all fields

skateman:
I tried to create the least invasive solution to add a `*` to the labels of fields that have an `isRequired: true` attribute. Most of the components are using the `prepareProps` in their `useFieldApi` call, so I added a `buildLabel` function in the same file. However, some components like `FieldArray` don't leverage the `prepareProps` so I am also exporting the `buildLabel` function and calling it explicitly where it is necessary. 

Screenshot from 2020-10-21 14-43-14

@kavyanekkalapu can you please verify if it's visually correct to have the `*` like this? The color is probably wrong and I will need a better one for sure.

### DIFF
--- a/packages/carbon-component-mapper/src/common/is-required.js
+++ b/packages/carbon-component-mapper/src/common/is-required.js
@@ -1,0 +1,19 @@
+import React from 'react';
+import { childrenPropTypes } from '@data-driven-forms/common/src/prop-types-templates';
+
+import './is-required.scss';
+
+const IsRequired = ({ children }) => (
+  <React.Fragment>
+    <span className="ddorg__carbon-component-mapper_is-required" aria-hidden="true">
+      *
+    </span>
+    {children}
+  </React.Fragment>
+);
+
+IsRequired.propTypes = {
+  children: childrenPropTypes
+};
+
+export default IsRequired;
--- a/packages/carbon-component-mapper/src/common/prepare-props.js
+++ b/packages/carbon-component-mapper/src/common/prepare-props.js
@@ -1,12 +1,15 @@
 import React from 'react';
 import WithDescription from './with-description';
+import IsRequired from './is-required';
 
-const prepareProps = ({ isDisabled, isReadOnly, label, description, ...props }) => ({
+export const buildLabel = (label, isRequired) => (label && (isRequired ? <IsRequired>{label}</IsRequired> : label)) || undefined;
+
+const prepareProps = ({ isDisabled, isReadOnly, isRequired = false, label, description, ...props }) => ({
   disabled: isDisabled,
-  labelText: label,
+  labelText: buildLabel(label, isRequired),
   readOnly: isReadOnly,
   ...props,
-  ...(description ? { labelText: <WithDescription description={description} labelText={label || props.labelText} /> } : {})
+  ...(description ? { labelText: <WithDescription description={description} labelText={buildLabel(label || props.labelText, isRequired)} /> } : {})
 });
 
 export default prepareProps;
--- a/packages/carbon-component-mapper/src/files/date-picker.js
+++ b/packages/carbon-component-mapper/src/files/date-picker.js
@@ -24,6 +24,7 @@ const DatePicker = (props) => {
 
 DatePicker.propTypes = {
   isDisabled: PropTypes.bool,
+  isRequired: PropTypes.bool,
   datePickerType: PropTypes.string,
   DatePickerProps: PropTypes.object,
   WrapperProps: PropTypes.object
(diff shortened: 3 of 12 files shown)


Hyperkid123:
@skateman so does this work with the official (optional) carbon guideline? Is it still there is you don't use the `isRequired` (sorry I do not know the carbon mapper implementation at all).

Hyperkid123:
```bash

/home/circleci/react-forms/packages/carbon-component-mapper/src/files/date-picker.js
  35:20  error  Delete `,`  prettier/prettier

/home/circleci/react-forms/packages/carbon-component-mapper/src/files/select.js
    2:21  error  'bool' is defined but never used  no-unused-vars
  211:20  error  Delete `,`                        prettier/prettier

```

skateman:
Well, if you don't set the `isRequired`, it is working according to the guidelines. If you set the `isRequired` there will be an additional `*` in the label, which I might wrap arund with a `<span>` that has a pretty class (working on it)...

kavyanekkalapu:
@skateman We can use color: #E0182D for * beside label, we need to specify on top of page with `* indicates a required field`( this is needed for A11y), i am just attaching one of the screenshots for sample.

Screen Shot 2020-10-21 at 8 52 28 AM

skateman:
The `* indicates a required field` hint can be added in our codebase globally, here I'm just setting the `*`:
Screenshot from 2020-10-21 15-00-18

packages/carbon-component-mapper/src/common/is-required.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { childrenPropTypes } from '@data-driven-forms/common/src/prop-types-templates';
+
+import './style.scss';
```
rvsia, line 4: I would call this file same as this component: `is-required.scss`

packages/carbon-component-mapper/src/files/date-picker.js
```diff
@@ -24,13 +24,15 @@ const DatePicker = (props) => {
...
+  isRequired: PropTypes.bool,
   datePickerType: PropTypes.string,
   DatePickerProps: PropTypes.object,
   WrapperProps: PropTypes.object
 };
 
 DatePicker.defaultProps = {
-  datePickerType: 'single'
+  datePickerType: 'single',
+  isRequired: false
```
rvsia, line 35: We don't need to set up `false` as a default value. By default, `undefined` is set so it works the same.

packages/carbon-component-mapper/src/common/prepare-props.js
```diff
@@ -1,12 +1,16 @@
 import React from 'react';
 import WithDescription from './with-description';
+import IsRequired from './is-required';
 
-const prepareProps = ({ isDisabled, isReadOnly, label, description, ...props }) => ({
+const buildLabel = (label, isRequired) => (label && (isRequired ? <IsRequired>{label}</IsRequired> : label)) || undefined;
```
rvsia, line 5: `export const buildLabel = ...`

packages/carbon-component-mapper/src/common/prepare-props.js
```diff
@@ -1,12 +1,16 @@
...
   disabled: isDisabled,
-  labelText: label,
+  labelText: buildLabel(label, isRequired),
   readOnly: isReadOnly,
   ...props,
-  ...(description ? { labelText: <WithDescription description={description} labelText={label || props.labelText} /> } : {})
+  ...(description ? { labelText: <WithDescription description={description} labelText={buildLabel(label || props.labelText, isRequired)} /> } : {})
 });
 
+export { buildLabel };
```
rvsia, line 15: ^^

packages/carbon-component-mapper/src/files/select.js
```diff
@@ -158,8 +160,9 @@ ClearedSelect.propTypes = {
   carbonLabel: PropTypes.node,
   placeholder: PropTypes.node,
   isDisabled: PropTypes.bool,
-  isSearchable: bool,
-  isClearable: bool
+  isRequired: PropTypes.bool,
+  isSearchable: PropTypes.bool,
+  isClearable: PropTypes.bool
```
rvsia, line 165: awesome, thanks! :+1:

rvsia (commented):
Small nitpicks, great job!

packages/carbon-component-mapper/src/common/is-required.js
skateman: I was inspired by the ant mapper, where it is done the same way. Should I really be inconsistent?

rvsia: So it's wrong there. :smile:

rvsia (approved):

rvsia:
@skateman  Oh, I forgot about the documentation.... Could you please also delete `no IsRequired` section from the documentation? https://data-driven-forms.org/mappers/carbon-component-mapper#noisrequired :pray:

PR #847, opened 2020-10-21
Add missing dual list select into the carbon demo schema

skateman:
There was no demo for the dual-list-select component for the Carbon mapper, so I'm adding one.


Hyperkid123 (approved):

Hyperkid123:
@skateman this should not trigger a new release. Can you remove the `fix(xxx):` from the commit message?

skateman:
> @skateman this should not trigger a new release. Can you remove the `fix(xxx):` from the commit message?

sure thing :+1:

rvsia (approved):

PR #850, opened 2020-10-22
Add .nvmrc into the gitignore

skateman:


Hyperkid123:
@skateman we actually require .nvmrc for releases. What is the motivation behind this change?

Hyperkid123:
NVM i thought it is npmrc not nvmrc...

Hyperkid123 (approved):

PR #851, opened 2020-10-22
feat(carbon): Implement searchable/clearable non-multi select

skateman:
I found a way to implement the `isSearchable` and `isClearable` functionalities for non-multi selects in carbon usin the [`ComboBox`](https://react.carbondesignsystem.com/?path=/docs/combobox--combobox) component. Unfortunately, these two features come hand in hand, it's not possible to turn off one just of them without turning off the other one as well. Therefore, I adjusted the logic to turn on both features if at least one of them is enabled, i.e. `isClearable || isSearchable` for both the single and the multiselect.

Screenshot from 2020-10-22 10-20-04

I guess I'll need some help with tests :pray: :see_no_evil:

### DIFF
--- a/packages/carbon-component-mapper/src/files/select.js
+++ b/packages/carbon-component-mapper/src/files/select.js
@@ -5,7 +5,7 @@ import { useFieldApi } from '@data-driven-forms/react-form-renderer';
 import DataDrivenSelect from '@data-driven-forms/common/src/select';
 import fnToString from '@data-driven-forms/common/src/utils/fn-to-string';
 
-import { Select as CarbonSelect, MultiSelect, SelectItem } from 'carbon-components-react';
+import { Select as CarbonSelect, MultiSelect, SelectItem, ComboBox } from 'carbon-components-react';
 import prepareProps from '../common/prepare-props';
 
 export const multiOnChange = (input, simpleValue) => ({ selectedItems }) => {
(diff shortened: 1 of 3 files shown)


Hyperkid123:
@skateman well for the tests you just really have to know the DOM structure and which elements trigger which events.

For the clearable feature I can image the tests would be something like this:

```jsx
// I dont know the DOM for the carbon combobox. Slectors and props are made up.
const wrapper = mount(<Select {...props} value="some value") />
expect(wrapper.find('ValueContainer').test()).toEqual("some value")

wrapper.find('.clear-button').prop('onClick')()
expect(wrapper.find('ValueContainer').test()).toEqual(undefined)
```

skateman:
@Hyperkid123 thinking your idea through, I don't think we need to test the feature of the `ComboBox` that is already verified by the tests in an external repo. Would it be enough to test if the right component is being rendered? You're doing the same kind of test for the single/multi selects already.

Hyperkid123:
@skateman sure if its carbon internal method it is enough. Just make sure that any action that is DFF specific is tested.

skateman:
> @skateman sure if its carbon internal method it is enough. Just make sure that any action that is DFF specific is tested.

I guess all I added is tested...

Hyperkid123 (approved):
Codecov agrees.

PR #869, opened 2020-10-29
Make dual-list pf4 mapper elements customizable

fhlavac:
Added necessary props to make dual-list pf4 mapper elements customizable

@Hyperkid123

### DIFF
--- a/packages/pf4-component-mapper/src/files/dual-list-select.d.ts
+++ b/packages/pf4-component-mapper/src/files/dual-list-select.d.ts
@@ -1,6 +1,8 @@
 import { UseFieldApiComponentConfig, AnyObject } from "@data-driven-forms/react-form-renderer";
 import { ReactNode } from "react";
 import FormGroupProps from "./form-group";
+import { FormGroupProps as FormGrpProps, ListProps, ToolbarProps, TextInputProps, ButtonProps, GridProps, GridItemProps, TextContentProps, FlexProps, FlexItemProps } from "@patternfly/react-core";
+import { SVGIconProps } from "@patternfly/react-icons/dist/js/createIcon";
 
 export interface DualListSelectOption extends AnyObject {
   value?: any;
@@ -16,6 +18,14 @@ export interface RenderStatusFunction {
   (input: RenderStatusInput): ReactNode;
 }
 
+export interface VoidFunction {
+  (): void;
+}
+
+export interface HandleOptionsFunction {
+  (e: any, value: any): void;
+}
+
 interface InternalDualListSelectProps {
   leftTitle?: ReactNode;
   rightTitle?: ReactNode;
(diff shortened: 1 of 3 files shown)


Hyperkid123:
@fhlavac I have renamed this to WIP. Once you have time, make all elements customizable like in other mappers. :bow:

Hyperkid123 (commented):
Good job. Now you need to add these props to the typescript module definition and to the documentation.

The typings file is called `dual-list-select.d.ts` and all the definitions must reflect their source. So if target element for a props object is `div` it has to use react div typings. If its for a PF `Toolbar` it has to use PF ToolbarProps definitione etc. You can find example in other components/mappers.

The documentation md file is: `packages/react-renderer-demo/src/doc-components/examples-texts/pf4/dual-list.md`. It should be called `dual-list-select.md` so rename it please and add all the new props to it with corresponding links to original component docs if possible. You can look at the MUI variant for guidance.

rvsia (commented):
As @Hyperkid123 d.ts field should be updated.

Also, you should update the documentation page for this component.

Otherwise, looks good!

packages/pf4-component-mapper/src/files/dual-list-select.js
```diff
@@ -53,11 +57,11 @@ const List = ({ value, optionClick, noOptionsTitle, filterValue, filterValueText
           onClick={(e) => optionClick(e, value)}
           key={value}
           value={value}
+          {...ListItemProps}
+          {...ListItemPropsItem}
           className={`ddorg__pf4-component-mapper__dual-list-select-option ${
             selectedValues.includes(value) ? 'ddorg__pf4-component-mapper__dual-list-select-option-selected' : ''
```
rvsia, line 63: You should convert this into clsx function and add `ListItemProps.className` and `ListItemPropsItem.className`

packages/pf4-component-mapper/src/files/dual-list-select.d.ts
```diff
@@ -25,15 +35,81 @@ interface InternalDualListSelectProps {
...
+  description?: ReactNode;
+  hideLabel?: boolean;
+  id?: string;
+  state?: AnyObject,
+  sortOptions?: VoidFunction,
+  filterOptions?: VoidFunction,
+  leftValues?: DualListSelectOption[];
+  rightValues?: DualListSelectOption[];
   renderStatus?: RenderStatusFunction;
+  handleOptionsClick?: HandleOptionsFunction,
```
Hyperkid123, line 56: all of these handlers, sorter, etc, should not be part of the interface. These functions are inject from the common dual list select implementation imported from
```jsx
import DualListSelectCommon from '../../../common/src/dual-list-select';
```
and used at
```jsx
const DualListSelectWrapper = (props) => <DualListSelectCommon {...props} DualListSelect={DualList} />;
```
You can't rewrite those from the schema. Only props configurable from schema should be listed here.

I think any the stuff after the `// props` comment in the prop types definition.

packages/react-renderer-demo/src/doc-components/examples-texts/pf4/dual-list-select.md
```diff
@@ -7,6 +7,7 @@ Dual list select is wrapped in a form group, so it accepts all [form group props
 |Props|Type|Default|Description|
 |-----|----|-------|-----------|
 |options|array|[]|`[{label, value}]`|
+|label|node|FormLabel primary text|
```
Hyperkid123, line 10: I think you are missing one `|` between FormLabel and primary text.

Hyperkid123 (approved):

rvsia:
Oh, there is no semantic-release commit message 😿

Hyperkid123:
:facepalm: does not mapper. It will be released at some point. :D I don't think we need it atm.

PR #881, opened 2020-11-05
feat(manager): focus on error input

### LINKED ISSUE #723: [manager] Add focus decorator

rvsia:
closes #723 

focus

### DIFF
--- a/packages/form-state-manager/demo/index.tsx
+++ b/packages/form-state-manager/demo/index.tsx
@@ -2,10 +2,13 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 
 import FormStateManager from '../src/files/form-state-manager';
+import FormSpy from '../src/files/form-spy';
+
 import TextField from '../src/tests/helpers/text-field';
 import { Validator } from '../src/types/validate';
 
-const asyncValidator: Validator = (value) => new Promise((res, rej) => setTimeout(() => (value === 'foo' ? rej('No async foo') : res()), 250));
+const asyncValidator: Validator = (value) => new Promise((res, rej) => setTimeout(() => (value === 'foo' ? rej('No async foo') : res()), 1000));
+const asyncValidator1: Validator = (value) => new Promise((res, rej) => setTimeout(() => (value === 'foo' ? rej('No async foo') : res()), 3000));
 
 const App = () => {
   return (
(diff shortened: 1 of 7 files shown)


packages/form-state-manager/src/utils/focus-error.ts
```diff
@@ -0,0 +1,17 @@
+import AnyObject from '../types/any-object';
+
+const focusError = (errors: AnyObject) => {
+  if (document) {
+    const allInputs = Array.from(document.forms)
```
Hyperkid123, line 5: this might get messy when multiple forms are on. Can we pass the exact form element to the state manager? We could use ref for the react wrapper form component.

rvsia: I don’t think so, because the form element is not created by the manager but in the render function. So we had to create a ref and then pass it in the render function to the form - and it would tied this library to React. 

But we could easily use name attribute, but that would need users to set the name on form element and pass the same name to form manager if they run into this issue. What do you think?

Otherwise, it works the same as FF now. 😄

Hyperkid123: Well, I don't think this is good. This can interfere with search inputs for instance. It's fairly common to have your search input wrapped in a form. Or any form that has input with the same name. I think we need some unique attribute (id or data) to distinguish between forms and not focus incorrect input.

rvsia: I am not sure what you mean.

You can put a name on form and the just find form with this name. It won’t interfere with any of inputs.

Hyperkid123: Yeah, I meant the current solution, your suggestion would work.

TLDR, we need to uniquely specify the form instance to prevent potential focus conflicts between multiple forms. I believe you will come up with an adequate solution.

packages/form-state-manager/demo/index.tsx
```diff
@@ -16,7 +16,7 @@ const App = () => {
       <FormStateManager onSubmit={console.log}>
         {({ handleSubmit, ...state }) => {
           return (
-            <form onSubmit={handleSubmit}>
+            <form onSubmit={handleSubmit} name="field-2">
```
Hyperkid123, line 19: Do you think it's feasible to add some generated name as a prop from `FormStateManager`? That way we can always have some name?

We can go even further and try to inject it into the first form of the child element. I know it may seem a bit too much but this could realle be an annoying bug.

rvsia: I wouldn’t do this, I don’t think it’s a huge bug - the worst thing that can happen is that after a failed submit you will be focused on a wrong input under these circumstances:

1. Having at least two forms with two inputs of the same name
2. Having the correct form as the first in dom
3. Have an error on the field with the same name
4. Do not disable buttons on invalid

With clear instructions on documenation it is simple to fix it. (And I think that even junior skilled developers will understand why it’s happening, as they see that the focused field has the same name.)

rvsia: To be clear, you don’t need to put a name there. It’s not required and it still works well. In the demo I just did it for testung purposes (to see what happens when a form and field has the same name)

Hyperkid123: All right. We can address it if it becomes a problem.

Hyperkid123 (approved):

PR #883, opened 2020-11-06
fix(all): update peer dependencies verion restrictions

Hyperkid123:


packages/ant-component-mapper/package.json
```diff
@@ -67,10 +67,10 @@
     "react-select": "^2.2.0"
   },
   "peerDependencies": {
-    "@data-driven-forms/react-form-renderer": "^1.9.3",
+    "@data-driven-forms/react-form-renderer": ">=1.9.3",
```
rvsia, line 70: I would change it to version 2, the same for mui mapper

rvsia (approved):

PR #889, opened 2020-11-16
fix(dual-list): set value to empty array when undefined

boaz0:
In dual list selector when the right side is empty, the value is set to undefined which causes that component to be discarded in the `values` object that is passed to `onSubmit`.

For example, I have three dual list selectors here:
image

If I click on "Submit", it will send an empty object {} as `values`.
With this patch it will send the following object:
```
{
 tags: {
  aws: { enabled: [] },
  azure: { enabled: [] },
  openshift: { enabled: [] }
 }
}
```

### DIFF
--- a/packages/blueprint-component-mapper/src/tests/dual-list-select.test.js
+++ b/packages/blueprint-component-mapper/src/tests/dual-list-select.test.js
@@ -254,7 +254,7 @@ describe('DualListSelect', () => {
       wrapper.find('form').simulate('submit');
     });
 
-    expect(onSubmit).toHaveBeenCalledWith({});
+    expect(onSubmit).toHaveBeenCalledWith({ 'dual-Menu': [] });
   });
 
   it('switch all to right', async () => {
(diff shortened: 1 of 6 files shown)


Hyperkid123:
@boaz0 I am i correct to assume that this is an issue when sending data to some update API endpoint?

Hyperkid123 (changes requested):
@boaz0 this is a breaking change for any other user of the dual list select component.

Can you check if [cleared value](https://data-driven-forms.org/schema/cleared-value#heading-clearedvalue) would help you instead?

If that is not the case, we would need some prop that to switch between the return value format.

boaz0:
It won't work unless there is a way to pass that through `schema`.

Hyperkid123:
@boaz0 you can use that config in schema directly. Or you can adjust the mapper and this prop globally specifically for the component via [global props](https://data-driven-forms.org/mappers/global-component-props#heading-globalcomponentprops).

boaz0:
OK, so the problem is:
1) I want to set value to `[]` if is `undefined`.
2) my front end gets the schema from the backend and that means I cannot write a JS in my schema.
Maybe we can add a prop to dual-list-select to use empty array?

Hyperkid123:
@boaz0 did the cleared value prop worked?

boaz0:
@Hyperkid123 yes it did. :+1:

PR #898, opened 2020-11-23
feat(manager): integrate manager with DDF

rvsia:
- [x] update tests
- :stop_sign:  render optimization - do separately
   - :stop_sign:  it seems that registering a lot fields at once is too slow :thinking:
   - :stop_sign: changing values is so slow (try to hold key on textinput)
- [x] subit focus already fixed errors (there is probably key with undefined)


Hyperkid123:
> it seems that registering a lot fields at once is too slow

I believe that's because is done one by one. We could do it in a batch.

Hyperkid123:
@rvsia what is the performance like?

rvsia:
@Hyperkid123  This is ready for review although:
- a lot of optimization needs to be done
- some PF3 tests are skipped because there will be removed anyways
- some new features are missing test coverage, will be done separately

Hyperkid123:
Ok we will take care of that in different PR

rvsia:
@Hyperkid123 Yeah, this is not intended to be the final PR, but just an initial integration to be able to test it in a real environment. After this is merged, the fun can start! :smile:

Hyperkid123 (approved):

PR #901, opened 2020-11-24
fix(mui): map component prop to element in plain text

rvsia:


Hyperkid123 (commented):
@rvsia can you map it directly to MUI type? https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Typography/Typography.js#L216

Also, we should exclude the `component` from TS typings and include the element.

Hyperkid123 (approved):

PR #905, opened 2020-11-27
Add custom components docs section.

Hyperkid123:
### Changes
- add custom components demo section to docs

cc @jinxac

### DIFF
--- a/packages/react-renderer-demo/src/components/navigation/schemas/schema.js
+++ b/packages/react-renderer-demo/src/components/navigation/schemas/schema.js
@@ -3,6 +3,7 @@ import schemaRenderer from './renderer.schema';
 import schemaNav from './schema.schema';
 import schemaHooks from './hooks.schema';
 import mappersSchema from './mappers.schema';
+import customExamplesSchema from './custom-examples.schema';
 
 const schema = [
   {
@@ -56,6 +57,12 @@ const schema = [
   {
     linkText: 'Optimization',
     link: 'optimization'
+  },
+  {
+    title: 'Examples',
+    link: 'examples',
+    noRoute: true,
+    fields: customExamplesSchema
   },
   {
     linkText: 'Releases',
--- a/packages/react-renderer-demo/src/components/navigation/schemas/custom-examples.schema.js
+++ b/packages/react-renderer-demo/src/components/navigation/schemas/custom-examples.schema.js
@@ -1,0 +1,8 @@
+const customExamplesSchema = [
+  {
+    component: 'sample-example',
+    linkText: 'Sample example'
+  }
+];
+
+export default customExamplesSchema;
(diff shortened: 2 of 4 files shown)


packages/react-renderer-demo/src/examples/components/custom-components/sample-example.js
```diff
@@ -0,0 +1,40 @@
...
+      component: 'custom-component',
+      name: 'some-name',
+      label: 'Custom component prop',
+      validate: [{ type: validatorTypes.REQUIRED, message: 'I am a required field' }]
+    }
+  ]
+};
+
+const SampleExample = () => {
+  return (
```
rvsia, line 33: We don't return block here

packages/react-renderer-demo/src/pages/custom-examples/sample-example.md
```diff
@@ -0,0 +1,20 @@
...
+
+
+# Sample example
+
+## Use markdown syntax to write text
+
+## What to include?
+- description of the component
+- what issue it can help solve
+- include component preview in a form
```
rvsia, line 14: Maybe I would use this page to say: 

"Hey if you need custom component, use useField and check this example"

We can put the text you used to `development setup` page. :thinking:

packages/react-renderer-demo/src/components/navigation/schemas/schema.js
```diff
@@ -57,6 +58,12 @@ const schema = [
     linkText: 'Optimization',
     link: 'optimization'
   },
+  {
+    title: 'Custom components demo',
```
rvsia, line 62: what about to remove `demo`?

rvsia (commented):
Just some suggestions.

One thing to do: after this is merged, update DocSearch config, there are defined entrypoints and we need to add each new menu subsection as an entrypoint to let algolia explore the menu.

packages/react-renderer-demo/src/components/navigation/schemas/schema.js
Hyperkid123: We already have custom components entry in the docs. I would like users to find the guide how to create the custom components before looking at this.

packages/react-renderer-demo/src/pages/custom-examples/sample-example.md
Hyperkid123: I fully expect this example to get deleted. This is here just for @jinxac so he knows where to pud his demo.

packages/react-renderer-demo/src/components/navigation/schemas/schema.js
rvsia: hmm so maybe 'Custom components examples'?

Hyperkid123: I think examples might be enough.

rvsia (approved):

PR #909, opened 2020-12-01
[WIP] New build process

Hyperkid123:
## Changes

This brings optimized build output for the render (other packages will follow).

### Build tool
We can ditch rollup and just use babel. There is no need for advanced bundling. Dependencies will be installed instead. This should reduce the overall build size by a lot.

### Import paths
Instead of directly pointing to `cjs`/`esm` builds we can now directly point to `@data-driven-forms/react-form-renderer/form-renderer`. That directory contains cjs build but also a separate `package.json`, which will resolve the correct import path either to cjs bundle or esm. 

The structure looks somethig like this
```
root/
├─ esm/
│  ├─ form-renderer/
│  │  ├─ index.js
│  │  ├─ form-renderer.js
├─ form-renderer/
│  ├─ package.json # makes the bundler to use esm module if it can and uses cjs otherwise. And typings
│  ├─ index.js
│  ├─ index.d.ts
│  ├─ form-renderer.js

```

todo
- [x] tests
- [x] docs (might be broken until all is migrated)
- [x] migrate all
  - [x] renderer
  - [x] mui mapper
  - [x] ant mapper
  - [x] blueprint mapper
  - [x] carbon mapper
  - [x] semantic mapper
  - [x] pf mapper (not pf3)
- [x] start releasing common package (this is required because we no longer bundle dependencies)
- [ ] figure out styling
- [x] test it with fresh project
- [x] make CI work again
- [ ] add typings to common package
- [x] create testing releases


rvsia:
I have created `3.0.0-beta.0` for mui, renderer, common and form state packages to test it.

rvsia (approved):
let's merge it now and fix styles later

PR #913, opened 2020-12-10
breaking(pf3): remove PF3 mapper

rvsia:
**Description**

Removes PF3 mapper. :skull: 

We can keep it on the documentation, but with changed warning - the warning is not there as it is currently only on master. I will update it in a follow-up pr where I will merge master to this branch.


Hyperkid123:
@rvsia no need to to keep it in docs. It will live in v2 branch and we can just move the MD files to the packages instead.

rvsia:
@Hyperkid123  removed from docs

Hyperkid123 (approved):

Issue #931, opened 2021-01-14
[V3] Refactor S/CSS to CSS-in-JS for version 3


rvsia:
**Scope**: All

**Description**

- [x] ant (css/scss) @rvsia https://github.com/data-driven-forms/react-forms/pull/937
- [x] blueprint @rvsia https://github.com/data-driven-forms/react-forms/pull/939
- [x] carbon @rvsia https://github.com/data-driven-forms/react-forms/pull/940
- [x] pf4 @rvsia https://github.com/data-driven-forms/react-forms/pull/942
- [x] geneate CSS in package files (PF4, ant) @rvsia https://github.com/data-driven-forms/react-forms/pull/968

PR #942, opened 2021-01-18
fix(pf4): change scss to css

rvsia:
- also removes obsolete schemas

Part of https://github.com/data-driven-forms/react-forms/issues/931


packages/pf4-component-mapper/package.json
```diff
@@ -64,8 +64,9 @@
   },
   "dependencies": {
     "@data-driven-forms/common": "*",
+    "downshift": "^5.4.3",
     "prop-types": "^15.7.2",
-    "downshift": "^5.4.3"
+    "react-jss": "^10.5.0"
```
Hyperkid123, line 69: this dependency is not necessary. You are using just CSS

rvsia: oh right, leftover

Hyperkid123 (approved):

PR #949, opened 2021-01-21
fix(carbon): pass the label for numeric text-field correctly

skateman:
The `TextInput` component takes a `labelText` prop, while the `NumberInput` needs a `label` :confused: :disappointed: 

**Before:**
Screenshot from 2021-01-21 16-37-23

**After:**
Screenshot from 2021-01-21 16-37-17

### DIFF
--- a/packages/carbon-component-mapper/src/files/text-field.js
+++ b/packages/carbon-component-mapper/src/files/text-field.js
@@ -7,7 +7,7 @@ import { TextInput, NumberInput } from 'carbon-components-react';
 import prepareProps from './prepare-props';
 
 const TextField = (props) => {
-  const { input, meta, validateOnMount, ...rest } = useFieldApi(prepareProps(props));
+  const { input, meta, validateOnMount, labelText, ...rest } = useFieldApi(prepareProps(props));
 
   const Component = input.type === 'number' ? NumberInput : TextInput;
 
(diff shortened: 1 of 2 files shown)


rvsia:
@skateman We have a test file to catch up these issues: https://github.com/data-driven-forms/react-forms/blob/master/packages/carbon-component-mapper/src/tests/components.test.js

Could you please update it with this new variant? 

1. add new component to component mapper under some key (`NumberInput`) and use [globalProps](https://data-driven-forms.org/mappers/global-component-props#heading-globalcomponentprops) to set up `type: 'number'`
2. add the key to the list
3. profit :moneybag:

packages/carbon-component-mapper/src/files/text-field.js
```diff
@@ -7,22 +7,37 @@ import { TextInput, NumberInput } from 'carbon-components-react';
...
-    <Component
+    <TextInput
       {...input}
       key={input.name}
       id={input.name}
       invalid={Boolean(invalid)}
       invalidText={invalid || ''}
       warn={Boolean(warn)}
       warnText={warn || ''}
+      labelText={labelText}
```
rvsia, line 40: I like the previous approach more. :thinking: 

What about to just change the label/labelText:

```jsx
{...(type === 'number' ? { labelText } : { label: labelText } )}
```

skateman:
@rvsia @Hyperkid123 can we get this in as it is or could you please help me with the tests?

Hyperkid123 (approved):
We will deal with the tests later. We don't have the capacity right now. So we can merge them and add them later on.

PR #957, opened 2021-02-18
fix(doc): Fix minor typo

nclavaud:
**Description**

Fix minor typo in schema doc.


Hyperkid123 (approved):
:heart:

rvsia (approved):

rvsia:
@nclavaud Hello and thanks for the PR!

Can you just please change the commit message - we don't use Semantic release for the documentation so there is no reason to use the specific message format (in the other case, it would release new versions of all packages). Changing it to `Fix minor typo in docs` would be sufficient. Thanks!

nclavaud:
Of course! Commit message has just been updated :-)

PR #962, opened 2021-02-24
fix(pf4): use pf modal for wizard

### LINKED ISSUE #961: [PF4] Use PF4 modal instead of custom one
**Scope**: PF4 mapper

**Description**

Our PF4 Wizard includes custom implementation of modal. We should use the patternfly one as the original wizard component.

rvsia:
Fixes #961 

**Description**

Using PF4 modal to wrap the modal.

**Schema**

```jsx
{
   component: 'wizard',
   inModal: true,
   ...
}
```


Hyperkid123 (approved):

Hyperkid123:
@rvsia can you include this PR as well? https://github.com/data-driven-forms/react-forms/pull/963

I think it's unnecessary to release two new versions just for the PF modal.

rvsia:
@Hyperkid123 They are two separate issues and just patch releases. 🤷

Hyperkid123:
Yeah I know but I would be happier if we had less releases

rvsia:
I think the number of releases is irrelevant. It's fully automated - and actually that's a reason why we are using semantic release, to not have opinions about releases. 😄

PR #964, opened 2021-02-24
Sync with master

rvsia:
Merges master to form-state-branch

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "private": true,
   "scripts": {
     "bootstrap": "lerna bootstrap",
-    "build": "yarn clean-build && lerna run build --concurrency 1 --ignore @data-driven-forms/react-renderer-demo --ignore @data-driven-forms/pf3-component-mapper",
+    "build": "yarn clean-build && lerna run build --concurrency 1 --ignore @data-driven-forms/react-renderer-demo",
     "vendor": "lerna run vendor",
     "test": "jest --coverage",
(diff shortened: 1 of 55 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   "private": true,
   "scripts": {
     "bootstrap": "lerna bootstrap",
-    "build": "yarn clean-build && lerna run build --concurrency 1 --ignore @data-driven-forms/react-renderer-demo --ignore @data-driven-forms/pf3-component-mapper",
+    "build": "yarn clean-build && lerna run build --concurrency 1 --ignore @data-driven-forms/react-renderer-demo",
```
Hyperkid123, line 11: We should ignore the PF3 mapper. The build was not adjusted and we won't continue with development.

package.json
```diff
@@ -28,8 +28,7 @@
   "jest": {
     "verbose": true,
     "testPathIgnorePatterns": [
-      "/node_modules/",
-      "/pf3-component-mapper/"
+      "/node_modules/"
```
Hyperkid123, line 31: ^^

package.json
rvsia: there is no pf3 mapper

Hyperkid123: Oh the package is gone? I forgot.

Hyperkid123 (approved):

package.json
rvsia: image

PR #970, opened 2021-03-02
Common typings

### LINKED ISSUE #934: [v3] Missing typings for common
**Scope**: v3/common

**Description**

Common package missing  typings

Also most of common subfolders in mapper is missing it too - but we should include them too.

Hyperkid123:
Fixes #934

Changes
- update lint to remove false-positive errors in TS files (main `declare const` errors)
- add common typings


packages/pf4-component-mapper/src/tests/wizard/wizard.test.js
```diff
@@ -203,6 +203,7 @@ describe('<Wizard />', () => {
   });
 
   it('should call enter handler when pressing enter', () => {
+    // eslint-disable-next-line no-import-assign
```
rvsia, line 206: Maybe we should set this rule for all testing eslint files.

Hyperkid123: It was used only in this file. I would like to keep it and use jest spyOn instead

rvsia (approved):

PR #972, opened 2021-03-02
Add migration guide to V3

rvsia:
Adds migration guide for V3

### DIFF
--- a/packages/react-renderer-demo/src/components/code-editor/index.js
+++ b/packages/react-renderer-demo/src/components/code-editor/index.js
@@ -48,7 +48,18 @@ const CodeEditor = ({ value, children, className, inExample }) => {
 
   const lang = className ? className.toLowerCase().replace('language-', '') : undefined;
   let content = value || children || '';
-  content = tranformImports(content);
+
+  // read props from code in --- { "key": value } ---\n format
+  let propsFromMD = content.match(/--- .* ---/);
+  if (propsFromMD) {
+    propsFromMD = JSON.parse(propsFromMD[0].replace(/-/g, ''));
+    content = content.replace(/--- .* ---\n/, '');
+  }
+
+  if (propsFromMD?.switchable !== false) {
+    content = tranformImports(content);
+  }
+
   content = content.substring(0, content.length - 1);
 
   return (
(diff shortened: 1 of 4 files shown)


packages/react-renderer-demo/src/pages/migration-guide-v3.md
```diff
@@ -0,0 +1,73 @@
...
+
+UMD format is no longer supported. The import paths for UMD packages now lead to CJS modules in testing environments and ESM modules in browser environments.
+
+### Changed import paths
+
+Due to the new build process, you have to change your imports path, if you use relative imports of specific components or if you consume the specific module system package. There are no longer module system and the dist folder specified.
+
+**Before**
+
+`import useFieldApi from '@data-driven-forms/react-form-renderer/dist/cjs/use-field-api';`
```
Hyperkid123, line 37: I think we could use the git diff syntax to better illustrate these changes:
```diff
-import useFieldApi from '@data-driven-forms/react-form-renderer/dist/cjs/use-field-api';
+import useFieldApi from '@data-driven-forms/react-form-renderer/use-field-api';`
```
I am not sure though if the current code component supports it

rvsia: Yeah.. I didn't use the editor because there is the automatic import paths changer. 😅 

But diff is supported, so I will reimplement the ability to turn off the import switching.

packages/react-renderer-demo/src/components/code-editor/index.js
```diff
@@ -48,7 +48,18 @@ const CodeEditor = ({ value, children, className, inExample }) => {
 
   const lang = className ? className.toLowerCase().replace('language-', '') : undefined;
   let content = value || children || '';
-  content = tranformImports(content);
+
+  // read props from code in --- { "key": value } ---\n format
+  let propsFromMD = content.match(/--- .* ---/);
```
Hyperkid123, line 53: Can we use the language to turn off the import transformation? Just disable it for diff?

rvsia: I would rather keep it - I also used it in the example in optimization page.

Hyperkid123: OK

Hyperkid123 (approved):

PR #977, opened 2021-03-09
feat(mui): allow customization for field array

rvsia:
This PR adds customization for field array.

Opening it as draft until we resolve our strategy for v2/v3 transition.


Hyperkid123 (commented):
Looks good. Can you add the props to the TS interface?

Hyperkid123 (approved):

PR #981, opened 2021-03-11
fix(renderer): do not use common to import PropTypes.node

rvsia:
Import common package to just import `export const childrenPropTypes = PropTypes.oneOfType([PropTypes.node, PropTypes.arrayOf(PropTypes.node)]);` ( === `PropTypes.node`) seems like a overkill. 🤔


Hyperkid123 (approved):

Hyperkid123:
@rvsia btw I think we should get rid of the common prop types.

Issue #932, opened 2021-01-14
[v3] Wrong testcoverage


rvsia:
**Scope**: v3

**Description**

v3 shows wrong coverage for wizard components. This alone decreases the coverage about 10%

rvsia:
Packages has wrongly set index.js so they took components from build (https://github.com/data-driven-forms/react-forms/issues/932)

And for some reason codecoverage is not able now to follow `common` package, maybe bacause it also takes the built version...

Hyperkid123:
@rvsia thats because the common packages is now actually a package. So the test files are using the source code, but the build bundle which then causes the test coverage to be reduced. We have to move the test files to the common package a run it from there.

PR #987, opened 2021-03-12
feat(renderer): introduce FieldProps

### LINKED ISSUE #986: Omit all FF props
**Scope**: renderer

**Description**

There are multiple props we do not catch in `useFieldApi`: parse, format, ... https://final-form.org/docs/react-final-form/types/FieldProps

However, we need to update the components that are using these props (format is used by some date pickers etc.)

rvsia:
Fixes #986 

**Description**

🚨  **BREAKING** 🚨 

## FieldProps introduction

A new [FieldProps](/schema/introduction#fieldprops) props is introduced. This object serves to pass values to Final Form [useField configuration](https://final-form.org/docs/react-final-form/types/FieldProps). Do not use for natively supported props (`initialValue`, `validate`, ...).

```diff
{
    name: 'component-with-complex-isEqual',
    component: 'dual-list-selector',
-   isEqual: (a, b) => isEqual(a, b),
+   FieldProps: {
+     isEqual: (a, b) => isEqual(a, b),
+   }
}
```

**Schema**

```jsx
{
  component: 'text-field',
  name: 'texttonumber',
  FieldProps: {
     parse: value => Number(Value)
  }
}
```

### DIFF
--- a/packages/common/src/dual-list-select/dual-list-select.js
+++ b/packages/common/src/dual-list-select/dual-list-select.js
@@ -33,7 +33,9 @@ const DualListSelectCommon = (props) => {
 
   const { DualListSelect, ...rest } = useFieldApi({
     ...props,
-    isEqual: (current, initial) => isEqual([...(current || [])].sort(), [...(initial || [])].sort())
+    FieldProps: {
+      isEqual: (current, initial) => isEqual([...(current || [])].sort(), [...(initial || [])].sort())
+    }
   });
 
   const leftValues = rest.options
--- a/packages/react-form-renderer/src/use-field-api/use-field-api.js
+++ b/packages/react-form-renderer/src/use-field-api/use-field-api.js
@@ -121,7 +121,11 @@ const useFieldApi = ({
     dataType,
     type: combinedProps.type,
     ...(Object.prototype.hasOwnProperty.call(combinedProps, 'initialValue')
-      ? {initialValue: combinedProps.initialValue}
+      ? { initialValue: combinedProps.initialValue }
+      : {}
+    ),
+    ...(Object.prototype.hasOwnProperty.call(combinedProps, 'value')
+      ? { value: combinedProps.value }
       : {}
     ),
     ...FieldProps,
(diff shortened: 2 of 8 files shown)


rvsia:
@Hyperkid123 please do not merge, I found an error with checkboxes

rvsia:
@Hyperkid123 yeah, the problem was with not passing `value` prop when used with radio or checkbox. I added the value to the configuration, it won't break anything and it will behave the same as before without any changes in code.

rvsia:
@Hyperkid123 but please give it a deep look as it has a potential to 🌏  💥

packages/react-form-renderer/src/use-field-api/use-field-api.js
```diff
@@ -10,7 +10,7 @@ import composeValidators from './compose-validators';
 import isEqual from 'lodash/isEqual';
 
 const calculateInitialValue = (props) => {
-  if (Object.prototype.hasOwnProperty.call(props, 'initialValue') && Object.prototype.hasOwnProperty.call(props, 'dataType')) {
+  if (Object.prototype.hasOwnProperty.call(props, 'initialValue') && props.dataType) {
```
Hyperkid123, line 13: Is there any reason why not to keep the `hasOwnProperty` check?

rvsia: Yeah, dataType has to be a string, so we don't have to check falsy values.

packages/common/src/dual-list-select/dual-list-select.js
```diff
@@ -33,7 +33,9 @@ const DualListSelectCommon = (props) => {
 
   const { DualListSelect, ...rest } = useFieldApi({
     ...props,
-    isEqual: (current, initial) => isEqual([...(current || [])].sort(), [...(initial || [])].sort())
+    FieldProps: {
+      isEqual: (current, initial) => isEqual([...(current || [])].sort(), [...(initial || [])].sort())
```
Hyperkid123, line 37: Can you add a test here to see if the default prop is properly used?

rvsia: Oh, this is common package - it isn't tested at all because of the build process, see #https://github.com/data-driven-forms/react-forms/issues/932

However, the components using it are all tested in their mappers.

Hyperkid123 (approved):
I've tested some schemas and mappers manually. Everything seems to be in order.

PR #998, opened 2021-03-18
Add example for one-row-layout

rvsia:
Adds an example for one-layout


packages/react-renderer-demo/src/pages/examples/one-row-layout.md
```diff
@@ -0,0 +1,36 @@
+import DocPage from '@docs/doc-page';
+import CodeExample from '@docs/code-example';
+
+<DocPage>
+
+# One row layout
+
+This example show how to place two fields in the same row. **Only for MUI mapper**.
```
Hyperkid123, line 8: I would add the MUI to the example title and link. That will make it exact.

Hyperkid123 (approved):

PR #1006, opened 2021-03-19
Validation rendering

Hyperkid123:


rvsia (approved):
great job!

Hyperkid123:
There are some places in the renderer where we could utilize the memo HOC, but I am not sure it will bring any significant performance update.

PR #1010, opened 2021-03-22
Use original DatePicker and TimePicker PF4 components

rvsia:
**Description**

Use original DatePicker and TimePicker PF4 components


packages/pf4-component-mapper/package.json
```diff
@@ -60,8 +60,8 @@
   },
   "peerDependencies": {
     "@data-driven-forms/react-form-renderer": ">=2.4.6",
```
Hyperkid123, line 62: We should also bump this to v3. In all packages not just here.

Hyperkid123 (approved):

PR #1014, opened 2021-03-23
fix(antd): antd switch should bind 'checked' not 'value'

roytan883:
fix(antd): Antd switch should bind 'input.checked' not 'input.value'.

**Description**
If not bind with 'input.checked', the initial value will always `false` no matter the real value you provide.

### DIFF
--- a/packages/ant-component-mapper/src/switch/switch.js
+++ b/packages/ant-component-mapper/src/switch/switch.js
@@ -23,7 +23,7 @@ export const Switch = (props) => {
     ...props,
     type: 'checkbox'
   });
-  const { name, value, onChange, onBlur } = input;
+  const { name, checked, onChange, onBlur } = input;
 
   return (
     <FormGroup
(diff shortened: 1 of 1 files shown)


Hyperkid123 (commented):
@roytan883 Thanks for the PR.
Can you adjust the commit message so its first line will use this format: `fix|(antd): description`. It's required for the releaser trigger.

roytan883:
@Hyperkid123  done

Hyperkid123:
@roytan883 it has to be the commit message not the PR title.

roytan883:
@Hyperkid123 done change commit message

Hyperkid123:
@roytan883 I think that we don't understand each other. We need the format in a commit message, not PR title or description. What I want you to do is to rename this commit: https://github.com/data-driven-forms/react-forms/pull/1014/commits/471a6449cd467ee46d98e1c20bc366ac96f66713 

You can achieve it by this terminal command
```bash
$ git commit --amend # opens text editor with the commit message
```

And add this string on the first line: `fix(antd): switch should bind 'checked' not 'value'`

roytan883:
@Hyperkid123  I did `git commit --amend` then `push -f origin master`. please check if it ok?

Hyperkid123:
@roytan883 yes, now it's perfect. Thanks!

Hyperkid123 (approved):

PR #1017, opened 2021-03-26
Optimize condition and field array performance

### LINKED ISSUE #1016: [RENDERER] Block unnecessary condition and hide field renders
**Scope**: Form renderer

**Description**
Based on this discussion: https://github.com/data-driven-forms/react-forms/discussions/1015#discussioncomment-532730 we have found some performance degradations in massive forms (500+ fields). Altough performance in forms like these will be a bit worse, there is still room for improvement.

This effort will reduce the number of renders for some components and look at why the `useFieldApi` hook is not working properly with subscriptions when using conditions and field arrays.

Thanks, @thecodejack for testing the limits of the renderer.

Hyperkid123:
fixes #1016 

**Description**
The issue was that any field wrapped in the condition is subject to render, even if the `when` value did not change. This introduced improved `FormSpy` that listens only on a set of values in the form state. The effect of the new spy is that we no longer trigger render cycle in every condition wrapped field and by that extension skips the condition evaluation because it's not required.

See profiler snapshots and examples in the linked issue.

### DIFF
--- a/packages/ant-component-mapper/package.json
+++ b/packages/ant-component-mapper/package.json
@@ -69,7 +69,8 @@
     "react-dom": ">=16.13.0"
   },
   "dependencies": {
-    "@data-driven-forms/common": "*"
+    "@data-driven-forms/common": "*",
+    "lodash": "^4.17.21"
   },
   "postpublish": "export RELEASE_DEMO=true"
 }
--- a/packages/react-form-renderer/src/form-renderer/form-renderer.js
+++ b/packages/react-form-renderer/src/form-renderer/form-renderer.js
@@ -26,8 +26,20 @@ const FormRenderer = ({
   ...props
 }) => {
   const [fileInputs, setFileInputs] = useState([]);
+  const registeredFields = useRef({});
   const focusDecorator = useRef(createFocusDecorator());
   let schemaError;
+
+  const setRegisteredFields = (fn => registeredFields.current = fn({...registeredFields.current}));
+  const internalRegisterField = (name) => {
+    setRegisteredFields(prev => prev[name] ? ({...prev, [name]: prev[name] + 1}) : ({...prev, [name]: 1}));
+  };
+
+  const internalUnRegisterField = (name) => {
+    setRegisteredFields(({[name]: currentField, ...prev}) => currentField && currentField > 1 ? ({[name]: currentField - 1, ...prev}) : prev);
+  };
+
+  const internalGetRegisteredFields = () => Object.entries(registeredFields.current).reduce((acc, [name, value]) => value > 0 ? [...acc, name] : acc, []);
 
   const validatorMapperMerged = { ...defaultValidatorMapper, ...validatorMapper };
 
@@ -80,8 +92,12 @@ const FormRenderer = ({
               reset,
               clearOnUnmount,
               renderForm,
+              internalRegisterField,
+              internalUnRegisterField,
               ...mutators,
-              ...form
+              ...form,
+              ffGetRegisteredFields: form.getRegisteredFields,
+              getRegisteredFields: internalGetRegisteredFields,
             }
           }}
         >
--- a/packages/react-form-renderer/src/form-renderer/render-form.js
+++ b/packages/react-form-renderer/src/form-renderer/render-form.js
@@ -1,9 +1,9 @@
 import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import set from 'lodash/set';
+import { Field } from 'react-final-form';
 import RendererContext from '../renderer-context';
 import Condition from '../condition';
-import { Field } from 'react-final-form';
 import getConditionTriggers from '../get-condition-triggers';
 
 const FormFieldHideWrapper = ({ hideField, children }) => (hideField ? <div hidden>{children}</div> : children);
@@ -50,7 +50,7 @@ const ConditionTriggerDetector = ({ values = {}, triggers = [], children, condit
           condition={condition}
           field={field}
         >
-        {children}
+          {children}
         </ConditionTriggerDetector>
       )}
     </Field>
@@ -161,8 +161,6 @@ SingleField.propTypes = {
   resolveProps: PropTypes.func
 };
 
-const renderForm = (fields) => {
-  return fields.map((field) => (Array.isArray(field) ? renderForm(field) : <SingleField key={field.name} {...field} />));
-};
+const renderForm = (fields) =>fields.map((field) => (Array.isArray(field) ? renderForm(field) : <SingleField key={field.name} {...field} />));
 
 export default renderForm;
--- a/packages/react-form-renderer/src/form-renderer/render-form.js
+++ b/packages/react-form-renderer/src/form-renderer/render-form.js
@@ -69,10 +69,10 @@ const FormConditionWrapper = ({ condition, children, field }) => {
   if (condition) {
     const triggers = getConditionTriggers(condition, field);
     return (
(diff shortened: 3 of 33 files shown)


Hyperkid123:
I have thought of one possible issue and that is we are not accepting the dot notation when constructing the reduced values object. Ill adjust the code.

rvsia:
As we discussed in DMs, there is the issue with an unwanted registering fields that are being created in the condition wrapper. We should make this setting opt-in and describe the caveat in the documentation, or we could update the getRegisteredFields to ignore fields created by conditions (however, the question is then how to know what fields are created by conditions or by the useFieldHook)

Or we can wait until our state manager is implemented - then it would be simple to implement FieldSpy. 😈

rvsia:
schema for checking the wizard

```jsx
const wizardSchema = {
  fields: [
    {
      component: componentTypes.WIZARD,
      name: 'wizzard',
      buttonLabels: {},
      title: 'Cokoliv',
      fields: [
        { name: 'first', nextStep: ({values: {path}}) => path ? 'path1' : 'path2', fields: [{
          name: 'path',
          component: componentTypes.SWITCH,
          label: 'Path 1?'
        }]},
        {
          name: 'path1',
          nextStep: 'summary',
          fields: [{
            name: 'text1',
            label: 'text1',
            validate: [{type: 'required'}]
          }]
        },
        {
          name: 'path2',
          nextStep: 'summary',
          fields: [{
            name: 'text2',
            label: 'text2',
            validate: [{type: 'required'}]
          }]
        },
        {
          name: 'summary',
          fields: [{
            name: 'summary',
            component: componentTypes.TEXTAREA,
            label: 'summary',
            condition: {
              or: [
                {when: 'text1', isNotEmpty: true},
                {when: 'text2', isNotEmpty: true},
              ]
            }
          }]
        }
      ]
    }
  ]
};
```

Hyperkid123:
@rvsia I've added a custom field registration method. The wizard works with the new registration. There I've kept the FF getRegisteredFields under an alias.

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -26,9 +26,20 @@ const FormRenderer = ({
   ...props
 }) => {
   const [fileInputs, setFileInputs] = useState([]);
+  const [registeredFields, setRegisteredFields] = useState({});
```
rvsia, line 29: What about to use ref to not trigger any renders?

Hyperkid123: I don't think this triggers any additional renders of the children :thinking: Let me check the profiler. But, using ref might cause some race conditions because we can't use a callback to override the value.

Hyperkid123: Oh yeah, it does :facepalm: we don't shield the render from parent re-renders. I'm just gonna slap the ref there and we will see what it does.

Hyperkid123: OK it almost totally undid all the work. We can mimic the callback from setState which will make using ref safer.

packages/react-form-renderer/src/form-renderer/render-form.js
```diff
@@ -15,18 +17,64 @@ FormFieldHideWrapper.defaultProps = {
...
+    return (
+      <ConditionTriggerWrapper condition={condition} values={values} field={field}>
+        {children}
+      </ConditionTriggerWrapper>
+    );
+  }
+
+  const name = internalTriggers.shift();
+  return (
+    <FieldProvider skipRegistration name={name} subscription={{ value: true }} render={({input: {value}}) => (
```
rvsia, line 45: I would still use the FF field - we don't need any of our useFieldApi features (and so we don't have to use any `skipRegistration`)

packages/react-form-renderer/src/form-renderer/render-form.js
```diff
@@ -111,6 +159,8 @@ SingleField.propTypes = {
   resolveProps: PropTypes.func
 };
 
-const renderForm = (fields) => fields.map((field) => (Array.isArray(field) ? renderForm(field) : <SingleField key={field.name} {...field} />));
+const renderForm = (fields) => {
+  return fields.map((field) => (Array.isArray(field) ? renderForm(field) : <SingleField key={field.name} {...field} />));
```
rvsia, line 163: return block

Hyperkid123:
@rvsia I think it's ready. Feel free to dive into the code.

packages/ant-component-mapper/package.json
```diff
@@ -69,7 +69,8 @@
     "react-dom": ">=16.13.0"
   },
   "dependencies": {
-    "@data-driven-forms/common": "*"
+    "@data-driven-forms/common": "*",
+    "lodash": "^4.17.21"
```
rvsia, line 73: Isn't lodash dependency of form-renderer?

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -26,9 +26,21 @@ const FormRenderer = ({
...
+  const internalRegisterField = (name) => {
+    setRegisteredFields(prev => prev[name] ? ({...prev, [name]: prev[name] + 1}) : ({...prev, [name]: 1}));
+  };
+
+  const internalUnRegisterField = (name) => {
+    setRegisteredFields(({[name]: currentField, ...prev}) => currentField && currentField > 1 ? ({[name]: currentField - 1, ...prev}) : prev);
+  };
+
+  const internalGetRegisteredFields = () => Object.entries(registeredFields.current).reduce((acc, [name, value]) => value > 0 ? [...acc, name] : acc, []);
+
```
rvsia, line 43: What about to do what I proposed (this seems unnecessary complex to me):

```jsx
  const registeredFields = useRef([]);
```

register: 
```jsx
const internalRegisterField = (name) => registeredFields.current.push(name)
```

unregister: 

```jsx
function removeElement(array, elem) {
    var index = array.indexOf(elem);
    if (index > -1) {
        array.splice(index, 1);
    }
}

const internalUnRegisterField = (name) => removeElement(registeredFields.current, name)
```

internalGetRegisteredFields

```jsx
const internalGetRegisteredFields = () => new Set(registeredFields.current)
```

packages/react-form-renderer/src/form-renderer/render-form.js
```diff
@@ -15,18 +17,66 @@ FormFieldHideWrapper.defaultProps = {
...
+};
+
+const FormConditionWrapper = ({ condition, children, field }) => {
+  if (condition) {
+    const triggers = getConditionTriggers(condition, field);
+    return (
+    <ConditionTriggerDetector triggers={triggers} condition={condition} field={field}>
+      {children}
+    </ConditionTriggerDetector>
+  );
```
rvsia, line 75: strange formatting here

packages/react-form-renderer/src/form-renderer/form-renderer.js
Hyperkid123: There is a couple of potential issues. First, I don't like the idea of having the same item multiple times inside the array. Using an object is slightly more resource-efficient than looking for indexes.

Another problem might be the usage of `push` which can create an out-of-sync structure when we re-create the array (splice in removeElement). Theoretically, we can push into an array that no longer exists.

Also this:
```jsx
const internalGetRegisteredFields = () => new Set(registeredFields.current)
```

Would return an instance of a set instead of an array. We would need either `Array.from` or `[...x]` destructor to actually get an array to preserve current API. So we need some value transformation anyway.

PR #1018, opened 2021-03-31
feat(pf4): introduce dual list tree select component

rvsia:
**Description**

Introduces support for isTree variant of Dual List Selector (but it doesn't work properly because of issues in PF4...)

**Schema** *(if applicable)*

```jsx
import { componentTypes as components } from '@data-driven-forms/react-form-renderer';

const output = {
  title: 'Testing dual list',
  description: 'Description of testing dual list',
  fields: [
    {
      component: components.DUAL_LIST_SELECT,
      name: 'dual-list-select',
      label: 'choose favorite animal',
      isTree: true,
      isSearchable: true,
      isSortable: true,
      options: [
        {
          value: 'cats',
          label: 'cats'
        },
        {
          value: 'cats_1',
          label: 'cats_1'
        },
        {
          value: 'cats_2',
          label: 'cats_2'
        },
        {
          value: 'zebras',
          label: 'zebras'
        },
        {
          value: 'pigeons',
          label: 'pigeons'
        }
      ],
      options1: [
        {label: 'bb', hasBadge: true, badgeProps: { isRead: true }, children: [
          {value: 'value1', label: 'c'},
          {value: 'value2', label: 'a'},
          //{label: 'Folder 2', children: [{label: 'value 3', value: 'value 3'}]}
        ]},
        {label: 'aaa', hasBadge: true, badgeProps: { isRead: true }, children: [
          {value: 'value12', label: 'a'},
          {value: 'value23', label: 'z'},
          {value: 'value231', label: 't'},
          {label: 'Folder 2', children: [{label: 'xsd', value: 'zsadas'}]}
        ]}
      ]
    }
  ]
};

export default output;
```

### DIFF
--- a/packages/react-renderer-demo/src/doc-components/examples-texts/pf4/dual-list-select.md
+++ b/packages/react-renderer-demo/src/doc-components/examples-texts/pf4/dual-list-select.md
@@ -135,3 +135,7 @@ const CustomRightSortButton = () => {
   );
 };
 ```
+
+### isTree
+
+`isTree` follows the same structure as the [original prop](https://www.patternfly.org/v4/components/dual-list-selector#duallistselector), but you have to provide `value` for all leaves (not folders). You can also use `label` instead of `text` to preserve the same API as the standard variant. There is no `getValueFromNode` option for the tree variant.
--- a/packages/pf4-component-mapper/src/dual-list-select/dual-list-select.js
+++ b/packages/pf4-component-mapper/src/dual-list-select/dual-list-select.js
@@ -4,6 +4,8 @@ import { DualListSelector } from '@patternfly/react-core';
 import { useFieldApi } from '@data-driven-forms/react-form-renderer';
 import isEqual from 'lodash/isEqual';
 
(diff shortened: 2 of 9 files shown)


rvsia:
@Hyperkid123 I created an issue https://github.com/data-driven-forms/react-forms/pull/1018

We can merge this, there will be no changes (I hope) when a fix is released. And if anyone wants to use this, they can, simple cases could work OK.

packages/react-renderer-demo/src/doc-components/examples-texts/pf4/dual-list-select.md
```diff
@@ -135,3 +135,7 @@ const CustomRightSortButton = () => {
   );
 };
 ```
+
+### isTree
+
+`isTree` follows the same structure as the original prop, but you have to provide `value` for all leaves (not folders). You can also use `label` instead of `text` to preserve the same API as the standard variant. There is no `getValueFromNode` option for the tree variant.
```
Hyperkid123, line 141: Can we add a link to PF docs example?

Hyperkid123 (approved):

PR #1023, opened 2021-04-06
fix(carbon): combobox initial value label

### LINKED ISSUE #1022: initialValues Display Incorrectly in Select Component with isSearchable Property
**Description:** When trying to set the initialValues for a select component with the `isSearchable` property, the select treats the initialValues as the text to input into the field as opposed to the value of one of the options.

```
<FormRenderer
    schema={schema}
    initialValues={{ test: "test1" }}
    FormTemplate={FormTemplate}
    componentMapper={componentMapper}
    onSubmit={console.log}
/>
...
{
    component: "select",
    name: "test",
    id: "test",
    label: "Select",
    isSearchable: true,
    options: [
        {
          value: "test1",
          label: "This is test 1"
        },
        { value: "test2", label: "This is test 2" },
        { value: "test3", label: "This is test 3" },
        { value: "test4", label: "This is test 4" }
    ]
}
```

With `isSearchable`:

Without:

**Example:** https://stackblitz.com/edit/mtevuf-byqvrw?file=schema.js

Hyperkid123:
fixes #1022 

Will be backported to v2 after merge

### DIFF
--- a/packages/carbon-component-mapper/src/select/select.js
+++ b/packages/carbon-component-mapper/src/select/select.js
@@ -10,11 +10,14 @@ import prepareProps from '../prepare-props';
 
 export const multiOnChange = (input, simpleValue) => ({ selectedItem, selectedItems }) => {
   if (simpleValue) {
-    return input.onChange(selectedItems.map(({ value }) => value) || selectedItem.value);
+    return input.onChange(selectedItems?.map(({ value }) => value) || selectedItem.value);
   } else {
     return input.onChange(selectedItems || selectedItem);
   }
 };
+
+const getMultiValue = (value, options) =>
+  (Array.isArray(value) ? value : [value]).map((item) => (typeof item === 'object' ? item : options.find(({ value }) => value === item))) || [];
 
 const ClearedMultiSelectFilterable = ({
   invalidText,
@@ -44,7 +47,7 @@ const ClearedMultiSelectFilterable = ({
     invalid={Boolean(invalidText)}
     invalidText={invalidText}
     items={options}
-    initialSelectedItems={Array.isArray(rest.value) ? rest.value : []}
+    initialSelectedItems={getMultiValue(rest.value, options)}
   />
 );
 
@@ -95,7 +98,7 @@ const ClearedMultiSelect = ({
     invalid={Boolean(invalidText)}
     invalidText={invalidText}
     items={options}
-    initialSelectedItems={Array.isArray(rest.value) ? rest.value : []}
+    initialSelectedItems={getMultiValue(rest.value, options)}
   />
 );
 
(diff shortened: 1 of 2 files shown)


rvsia (commented):
Could you please also take a look on `isSearchable`+`simpleValue` variant? Now, combobox returns the whole option object. I am not sure if it is intended or not. 😬

Hyperkid123:
@rvsia id does not return the whole object, just the value. I don't even think simple value exists in carbon components. I believe it always returns just the value from the option.

Hyperkid123:
The simple value was OK. But all other select variants had same/similar issues with the initial value text. I've added accessors to properly create the initial selected prop.

rvsia:
@Hyperkid123 I see an error with this schema:

```jsx
            schema={{
              fields: [
                {
                  component: 'select',
                  name: 'test',
                  id: 'test',
                  label: 'Select',
                  isSearchable: true,
                  simpleValue: true,
                  options: [
                    {
                      value: 'test1',
                      label: 'This is test 1'
                    },
                    { value: 'test2', label: 'This is test 2' },
                    { value: 'test3', label: 'This is test 3' },
                    { value: 'test4', label: 'This is test 4' }
                  ]
                }
              ]
            }}
```

```jsx
select.js:13 Uncaught TypeError: Cannot read property 'map' of undefined
    at Object.onChange (select.js:13)
    at Object.onChange (ComboBox.js:144)
    at Downshift.<anonymous> (downshift.esm.js:681)
```


Kapture 2021-04-06 at 13 34 10

rvsia:
And it does return a whole object

```jsx
            schema={{
              fields: [
                {
                  component: 'select',
                  name: 'test',
                  id: 'test',
                  label: 'Select',
                  isSearchable: true,
                  //simpleValue: true,
                  options: [
                    {
                      value: 'test1',
                      label: 'This is test 1'
                    },
                    { value: 'test2', label: 'This is test 2' },
                    { value: 'test3', label: 'This is test 3' },
                    { value: 'test4', label: 'This is test 4' }
                  ]
                }
              ]
            }}
```

Kapture 2021-04-06 at 13 37 12

Hyperkid123:
> And it does return a whole object

@rvsia that is because the default `simpleValue` is set to false. It is supposed to return the whole object.

rvsia:
@Hyperkid123 Yeah, but normal select returns only the value. We need to make it consistent. (And in multi you can use simpleValue)

Hyperkid123:
I agree. But I can't tell you why it's here. At this point, it could break existing forms.

rvsia:
Yep and that's why we should enable to use simpleValue on isSearchable single select - it won't break anything.

Initially, there was no comboBox and the single select was always returning just the value.. but ComboBox broke this behavior and nobody catched that.

Hyperkid123:
I think we should address it in another issue. Now that I see the code, there are more issues with the component than just simple values, etc. It needs a major clean-up.

I've fixed the onChange crash.

rvsia (approved):

PR #1025, opened 2021-04-07
Implement FieldSpy

### LINKED ISSUE #1020: [V4] Implement FieldSpy
**Scope**: state manager

**Description**

With the latest changes in optimization of conditions, we have to introduce FieldSpy component that will listen changes on one field only.

rvsia:
Fixes #1020 

**Description**

Adds `FieldSpy` component.

Accepts these props:

`names` array of field names
`subscription`
`onChange`

### DIFF
--- a/packages/form-state-manager/src/field-spy/field-spy.ts
+++ b/packages/form-state-manager/src/field-spy/field-spy.ts
@@ -16,7 +16,7 @@ export interface FieldSpyProps {
   names: string[];
 }
 
-const FieldSpy: React.FunctionComponent<FieldSpyProps> = ({ names, children, onChange, subscription }) => {
+const FieldSpy: React.ComponentType<FieldSpyProps> = ({ names, children, onChange, subscription }) => {
   const { subscribe, unsubscribe, getState, formOptions, batch } = useContext(FormManagerContext);
   const [, rerender] = useReducer((prev) => prev + 1, 0);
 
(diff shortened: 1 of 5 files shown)


packages/form-state-manager/src/field-spy/field-spy.ts
```diff
@@ -0,0 +1,61 @@
...
+
+type OnChange = (managerState: ChildrenFieldSpyProps) => void;
+
+export interface FieldSpyProps {
+  subscription?: Subscription;
+  onChange?: OnChange;
+  names: string[];
+}
+
+const FieldSpy: React.FunctionComponent<FieldSpyProps> = ({ names, children, onChange, subscription }) => {
```
Hyperkid123, line 19: Can we use `React.ComponentType` instead of function component?

packages/form-state-manager/src/field-spy/field-spy.ts
```diff
@@ -0,0 +1,61 @@
...
+  );
+
+  const newState = { ...getState(), form: formOptions };
+
+  if (onChange) {
+    onChange(newState);
+    return null;
+  }
+
+  if (typeof children === 'function') {
```
Hyperkid123, line 54: If we are enforcing children to be function, we should also log/throw an error to let the devs know what to do.

Hyperkid123 (approved):

PR #1030, opened 2021-04-14
Update clean script to run only before build

rvsia:
We cannot run clean before install because we are missing packages needed in the clean script.


Hyperkid123 (approved):

Sahil2712 (approved):
This was causing an error, Thanks Richard for the changes.

PR #1036, opened 2021-04-16
Update stackblitz templates

rvsia:
The only issue that happens is this one:

Screenshot 2021-04-16 at 16 01 04

It does because of react-jss. 🤔  Trying to figure what is going there.


rvsia:
https://github.com/cssinjs/jss/issues/1329

rvsia:
We can merge this and fix react-jss issue after. This should at least fix PF4 examples...

Hyperkid123 (approved):

Issue #1037, opened 2021-04-16
Text Field with type: "number" treated as Empty when Incrementing/Decrementing


DavidResende0:
**Description**: When using a Text Field with the `type` property set to "number", the form will treat the field as if it's empty when entering a number by incrementing or decrementing. The form still recognizes manually entered numbers or numbers entered as an initialValue.

```
{
    component: "text-field",
    id: "number",
    isRequired: true,
    label: "Number Test",
    name: "number",
    step: 100,
    type: "number"
}
```

Manually entered number:
<img src="https://user-images.githubusercontent.com/64800041/115070346-d50a5400-9ec2-11eb-9974-d46dda2f6d32.png" width="300"> 

Incremented number:
<img src="https://user-images.githubusercontent.com/64800041/115070556-16026880-9ec3-11eb-9149-3f97f6f1c0ac.png" width="300"> 

**Schema**: https://stackblitz.com/edit/number-test?file=index.js

rvsia:
@DavidResende0 

This is an issue on Carbon side. Check it here: https://codesandbox.io/s/awesome-rhodes-4b161

When you click on any of the arrow button, `onChange` event is called with empty `target.value`.

There have already been multiple issues opened: https://github.com/carbon-design-system/carbon/issues?q=is%3Aissue+is%3Aopen+numberInput

So, I am closing this for now. Please, feel free to reopen, if you think we can do something with it here in DDF repo.

kavyanekkalapu:
@rvsia According to carbon docs, value is not present in e.target.value, it is in `e.imaginaryTarget.value` , check https://react.carbondesignsystem.com/?path=/docs/components-numberinput--default.

Example: https://codesandbox.io/s/misty-sea-yygnm?file=/src/index.js:463-486

Hyperkid123:
@kavyanekkalapu seems like an easy enough change. Can you open a PR?

kavyanekkalapu:
@Hyperkid123 Yes i can do that, but i don't see any code related to value in https://github.com/data-driven-forms/react-forms/blob/master/packages/carbon-component-mapper/src/text-field/text-field.js, so what do i need to do here, add a on change func like 
```if number input
value-->e.imaginaryTarget.value

rvsia:
Something like this: `onChange={e => input.onChange( ... )}`

kavyanekkalapu:
[...]

PR #1039, opened 2021-04-19
fix(carbon): textinput type Number onchange function

kavyanekkalapu:
**fixes:** https://github.com/data-driven-forms/react-forms/issues/1037

**Description**
For number input value should be e.imaginaryTarget.value instead of e.target.value

### DIFF
--- a/packages/carbon-component-mapper/src/tests/text-field.test.js
+++ b/packages/carbon-component-mapper/src/tests/text-field.test.js
@@ -1,0 +1,152 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import { act } from 'react-dom/test-utils';
+
+import { FormRenderer, componentTypes } from '@data-driven-forms/react-form-renderer';
+
+import FormTemplate from '../form-template';
+import componentMapper from '../component-mapper';
+import { TextInput, NumberInput } from 'carbon-components-react';
+
+describe('<TextInput and NumberInput />', () => {
+  let onChange;
+  let wrapper;
+  let submitSpy;
+  beforeEach(() => {
+    submitSpy = jest.fn();
+    onChange = jest.fn();
+  });
+  afterEach(() => {
+    onChange.mockReset();
+    submitSpy.mockReset();
+  });
+  it('renders NumberInput', () => {
+    const schema = {
+      fields: [
+        {
+          component: componentTypes.TEXT_FIELD,
+          name: 'input',
+          label: 'Please enter a value',
+          type: 'number',
+          step: 1,
+          initialValue: 1
+        }
+      ]
+    };
+
+    const wrapper = mount(
+      <FormRenderer
+        onSubmit={(values) => submitSpy(values)}
+        FormTemplate={(props) => <FormTemplate {...props} />}
+        schema={schema}
+        componentMapper={componentMapper}
+      />
+    );
+
+    expect(wrapper.find(NumberInput)).toHaveLength(1);
+  });
+
+  it('NumberInput - click on increment button', async () => {
+    const schema = {
+      fields: [
+        {
+          component: componentTypes.TEXT_FIELD,
+          name: 'input',
+          label: 'Please enter a value',
+          type: 'number',
+          step: 1,
+          initialValue: 1
+        }
+      ]
+    };
+    await act(async () => {
+      wrapper = mount(
+        <FormRenderer
+          onSubmit={(values) => submitSpy(values)}
+          FormTemplate={(props) => <FormTemplate {...props} />}
+          schema={schema}
+          componentMapper={componentMapper}
+        />
+      );
+    });
+
+    expect(wrapper.find(NumberInput)).toHaveLength(1);
+    expect(wrapper.find('input')).toHaveLength(1);
+
+    await act(async () => {
+      wrapper
+        .find('button.up-icon')
+        .first()
+        .simulate('click');
+    });
+    await act(async () => {
+      wrapper.find('form').simulate('submit');
+    });
+    wrapper.update();
+    expect(submitSpy).toHaveBeenCalledWith({ input: '2' });
+  });
+
+  it('NumberInput - click on decrement button ', async () => {
+    const schema = {
+      fields: [
+        {
+          component: componentTypes.TEXT_FIELD,
+          name: 'input',
+          label: 'Please enter a value',
+          type: 'number',
+          step: 1,
+          initialValue: 5
+        }
+      ]
+    };
+    await act(async () => {
+      wrapper = mount(
+        <FormRenderer
+          onSubmit={(values) => submitSpy(values)}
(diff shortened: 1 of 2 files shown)


packages/carbon-component-mapper/src/text-field/text-field.js
```diff
@@ -23,6 +27,7 @@ const TextField = (props) => {
       invalidText={invalid || ''}
       warn={Boolean(warn)}
       warnText={warn || ''}
+      onChange={(e) => setValue(e, input)}
```
rvsia, line 30: you have to call `input.onChange(e)` somewhere to change the value in the form state. see https://data-driven-forms.org/hooks/use-field-api#onchange

packages/carbon-component-mapper/src/text-field/text-field.js
```diff
@@ -11,6 +11,10 @@ const TextField = (props) => {
 
   const Component = input.type === 'number' ? NumberInput : TextInput;
 
+  const setValue = (e, input) => {
+    input.value = input.type === 'number' ? e.imaginaryTarget.value : e.target.value;
```
rvsia, line 15: This won't work. `input` is immutable, you cannot rewrite. You have to call `onChange` to change it and trigger a new render.

kavyanekkalapu: @rvsia this code is working in my local, does this work?
```
 const setValue = (e, input) => {
    if (input.type === 'number') {
      return input.onChange(e.imaginaryTarget.value);
    } else {
      return input.onChange(e.target.value);
    }
  };

Comment (Hyperkid123) [reply packages/carbon-component-mapper/src/text-field/text-field.js:15]: @kavyanekkalapu it "works" because there are also other events triggering that are causing the re-render. If you start limiting the `subscription` prop to the field, it will stop working at some point.
The `onChange` function is expecting either an HTML event or a value not undefined.

The code snippet you posted should work. Although this would be closer to our code style (forgive any typos in code example I am writing this on my phone)

```jsx
const getValue = (event) => event.type === 'number' ? event.imaginaryTarget.value : event.target.value

...

<Component
  ...
  onChange={event => input.onChange(getValue(event))}
  {...(input.type === 'number' ? { label: labelText } : { labelText })}
  {...rest}
/>

```

I would also like to see a test case.

packages/carbon-component-mapper/src/tests/text-field.test.js
```diff
@@ -0,0 +1,132 @@
...
+
+import FormTemplate from '../form-template';
+import componentMapper from '../component-mapper';
+import { TextInput, NumberInput } from 'carbon-components-react';
+
+describe('<TextInput and NumberInput />', () => {
+  let onChange;
+  let wrapper;
+  beforeEach(() => {
+    onChange = jest.fn();
```
rvsia, line 15: `onChange` is not being used - and you should not override onChange in the schema if you have not a really special case.

packages/carbon-component-mapper/src/tests/text-field.test.js
```diff
@@ -0,0 +1,132 @@
...
+      wrapper
+        .find('button.up-icon')
+        .first()
+        .simulate('click');
+    });
+    wrapper.update();
+
+    const valueInput = wrapper.find('input[type="number"]');
+    expect(valueInput).toHaveLength(1);
+    expect(valueInput.props().value).toEqual(2);
```
rvsia, line 75: We should test form state, not the value prop - the value prop can came up from the carbon component, so it can just “look” ok.

You can check example [here](https://data-driven-forms.org/testing#testingtherenderer). Just put a spy function to renderer submit and submit the form via `find(“form”).simulate(“submit”)` in `act` and check what values are received in the first argument.

kavyanekkalapu: @rvsia i tried it write it the way you mentioned, i am getting error,  not really sure what i missed, any idea?
Screen Shot 2021-04-19 at 2 13 21 PM

Screen Shot 2021-04-19 at 2 13 26 PM

Screen Shot 2021-04-19 at 2 14 22 PM

rvsia: Did you pass the spy function to renderer?

kavyanekkalapu: @rvsia your guess is correct, i haven't passed that :). updated it,  now getting new error
Screen Shot 2021-04-19 at 2 31 21 PM

rvsia: This is correct, there is formApi as the second argument in the submit function and some callback as the third. You don’t have to check these ones.

You can pick just the values in the submit prop:

`onSubmit={values => onSubmit(values)}`

kavyanekkalapu: @rvsia ah ok makes sense, i updated it now, i am still having issue with expected values, i am simulating clicks on increment and decrement buttons here, but on submit, it is still giving out initial value, i pushed my changes , please check.
Note: code i wrote for value check , it is working as expected.

rvsia: Yeah, it’s important to send the correct values in the submit.. I will check tomorrow morning what is going on, getting sleepy. 😴 Thank you for now! 👍

kavyanekkalapu: Cool, thank you.

packages/carbon-component-mapper/src/tests/text-field.test.js
```diff
@@ -0,0 +1,178 @@
...
+    const schema = {
+      fields: [
+        {
+          component: componentTypes.TEXT_FIELD,
+          name: 'input',
+          label: 'Please enter a value',
+          type: 'number',
+          step: 1,
+          initialValue: 1,
+          onChange
```
rvsia, line 33: Oh, this is the issue. You are still overwriting the onychange event.

kavyanekkalapu: ah ok. i will try to remove that and see, thanks.

kavyanekkalapu:
@rvsia I pushed the changes, thanks for helping me with testcases.

rvsia (approved):
Looks great! Thank you! 🏆

Hyperkid123 (commented):
@kavyanekkalapu can you please change the commit message format to trigger an automated release?. An example is here: https://github.com/data-driven-forms/react-forms#commits

kavyanekkalapu:
> @kavyanekkalapu can you please change the commit message format to trigger an automated release?. An example is here: https://github.com/data-driven-forms/react-forms#commits

Done, we need this bug for in 2.X.X release too, do i need to raise another pr for that?

Hyperkid123:
@kavyanekkalapu nope, we will backport it and trigger release.

PR #1043, opened 2021-04-21
feat(wizard): enable conditional submit step conversion

Hyperkid123:
based on a feature request in discord

If a conditional next step contains a submit flag (can be configured), the step will show submit button instead of a next button and wizard won't progress further.
```jsx
nextStep: {
        when: 'source-type',
        stepMapper: {
          aws: 'aws-step',
          google: CONDITIONAL_SUBMIT_FLAG,
          ...
        },
},
```

Flag value can be configured via `conditionalSubmitFlag` flag in wizard component schema:
```jsx
const schema = {
  fields: [{
    {
      "component": "wizard",
      "name": "wizard",
      // custom flag definition
      conditionalSubmitFlag: 'make-me-final-step',
      fields: [{
        name: 'step-1',
        // next step definition
        nextStep: {
          when: 'value',
          stepMapper: {
            "value-one": 'step-2',
            "value-two": 'make-me-final-step',
          },
        }
        ...
      }, {
        name: 'step-2'
        ...
      }]
...
  }}]
}
```

@rvsia if you can think of a better name/value for the flag, feel free to post it. I could not come with anything better.

### DIFF
--- a/packages/common/src/wizard/wizard.d.ts
+++ b/packages/common/src/wizard/wizard.d.ts
@@ -1,5 +1,4 @@
 import { AnyObject, Field } from "@data-driven-forms/react-form-renderer";
-
 export interface WizardProps extends AnyObject {
   fields: Field[];
   isDynamic?: boolean;
--- a/packages/react-renderer-demo/src/doc-components/examples-texts/wizard.md
+++ b/packages/react-renderer-demo/src/doc-components/examples-texts/wizard.md
@@ -1,10 +1,11 @@
 ## Docs for steps
 
-|Props|Type|Description|
-|----|-------------|----------------|
-|name|string,number|Name of the step|
-|nextStep|object/stepKey of next step/function|See below|
-|fields|array|As usual|
+|Props|Type|Required|Default|Description|
+|-----|----|--------|-------|-----------|
+|name|string,number|no|`undefined`|Name of the step|
+|nextStep|object/stepKey of next step/function|no|`undefined`|See below|
+|fields|array|yes|`undefined`|As usual|
+|conditionalSubmitFlag|string|no|`"@@ddf-common-wizard__conditional-submit-step"`|See below|
 
 ### nextStep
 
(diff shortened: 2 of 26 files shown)


packages/react-renderer-demo/src/doc-components/examples-texts/wizard.md
```diff
@@ -39,6 +40,58 @@ another option is to use custom function. The custom function receives as the fi
...
+Or with custom constant value defined by `conditionalSubmitFlag` prop:
+
+```jsx
+const schema = {
+  fields: [{
+    {
+      "component": "wizard",
+      "name": "wizard",
+      // custom flag definition
+      conditionalSubmitFlag: 'make-me-final-step',
```
rvsia, line 74: what about to call it: `submitStep`?

And what about to support also array of strings? 🤔

packages/common/src/wizard/wizard.d.ts
```diff
@@ -1,5 +1,6 @@
 import { AnyObject, Field } from "@data-driven-forms/react-form-renderer";
 
+export const CONDITIONAL_SUBMIT_FLAG = '@@ddf-common-wizard__conditional-submit-step';
```
rvsia, line 3: maybe it could be in a separate file so people do not have to import everything else in rare cases

Hyperkid123: If you use this constant without the common wizard, you would need to implement the whole wizard logic therefore this constant no longer serve any purpose.

rvsia: Yeah, but you can have just a schema where you want to use this value...

Or you can maybe implement your own wizard component but keep the constant the same.

packages/react-renderer-demo/src/doc-components/examples-texts/wizard.md
Hyperkid123: Maybe. What would be the use case? I am trying to think why would you need multiple flags for the same purpose. The only I can think of would be using two different wizards in a single form but that in itself does not make a whole lot of sense.

Or having two different flags in two different forms, but at that point, you have a unique flag per form.

rvsia: The question is "why not"? It would bring better flexibility and the code change is just changing `===` to `.includes()`

Hyperkid123: The why not is quite a simple answer for me. Because nobody has asked for it. There is a lot of "why not" features we are not supporting (remember schema parsers?).

packages/common/src/wizard/wizard.d.ts
Hyperkid123: Ill split it

packages/react-renderer-demo/src/doc-components/examples-texts/wizard.md
rvsia: Yeah, of course. If anyone would need this, we can change it,

rvsia (approved):

nathanbaleeta:
When considering option 1, I got an import error when I followed the instructions in the documentation [here](https://data-driven-forms.org/provided-mappers/wizard?mapper=mui).

This import doesn't work:
`import CONDITIONAL_SUBMIT_FLAG from '@data-driven-forms/common/wizard/c-o-n-d-i-t-i-o-n-a-l_-s-u-b-m-i-t_-f-l-a-g';`

This works:
`import { CONDITIONAL_SUBMIT_FLAG } from "@data-driven-forms/common/wizard";`

rvsia:
@nathanbaleeta  upps the import line is wrong in the documentation, thanks for reporting this! 👀

Hyperkid123:
Ah sorry about that, yes it's named export.

PR #1044, opened 2021-04-22
Do not translate wizard flag import

rvsia:
By default, we are using imports from root and editor automatically translates them. In this case we want to disable it.

cc @nathanbaleeta 

Screenshot 2021-04-22 at 8 34 58


Hyperkid123 (approved):
ofc

PR #1045, opened 2021-04-23
Add example for custom wizard

rvsia:
Adds an example for custom wizard component using `common/wizard` package. Just the most basic stuff, no fancy features.

### DIFF
--- a/packages/react-renderer-demo/src/examples/components/examples/custom-wizard.js
+++ b/packages/react-renderer-demo/src/examples/components/examples/custom-wizard.js
@@ -1,0 +1,117 @@
+import React, { useContext, useState } from 'react';
+
+import FormRenderer from '@data-driven-forms/react-form-renderer/form-renderer';
+import WizardContext from '@data-driven-forms/react-form-renderer/wizard-context';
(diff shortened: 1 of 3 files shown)


packages/react-renderer-demo/src/examples/components/examples/custom-wizard.js
```diff
@@ -0,0 +1,117 @@
...
+  const { formOptions, currentStep, handlePrev, onKeyDown, handleNext, activeStepIndex } = useContext(WizardContext);
+
+  return (
+    <div onKeyDown={onKeyDown} style={{ width: '100%' }}>
+      {currentStep.title}
+      {formOptions.renderForm(currentStep.fields)}
+      <FormSpy>
+        {() => (
+          <React.Fragment>
+            {currentStep.nextStep && (
```
Hyperkid123, line 79: We might want to use the same condition with the new final step flag for the next step resollution.

Hyperkid123 (approved):

packages/react-renderer-demo/src/examples/components/examples/custom-wizard.js
rvsia: No fancy stuff. 👁️

PR #1053, opened 2021-05-14
feat(mui-component-mapper)  upgrade to  mui v5

guyathomas:
**Description**
Upgrade MUI dep to v5 ( and as a result react 17 ) 

Not sure if this would be best incorporated in the next major version

- [x] `Yarn build` passes
- [x] `Yarn lint` passes
- [x] `Yarn test` passes
- [ ] Test coverage for new code *(if applicable)*
- [ ] Documentation update *(if applicable)*
- [x] Correct commit message
   - scope: package


rvsia:
Hello, thank you for this pr.

It failed on build: `error fs-extra@10.0.0: The engine "node" is incompatible with this module. Expected version ">=12". Got "10.24.1"` I am not sure if we can increase the node version because of the firebase hosting. 🤔 

Otherwise, if I see it correctly, MUI5 is still in the alpha release, so I think we should wait with this PR until the version 5 is stable. However, I don’t see any changes in the mui component mapper. If so, customers should be able to use the version 5 just by setting the dependency in their projects.

Do you have any particular reason for this pr?

guyathomas:
Hi! Thanks for the feedback. An interesting fact about the node version - I'll take it back down and see how we go.

The specific motivation for this was I'm working on a project with MUI5, and wanted to be able to test this lib out. Still very much WIP, and happy to close until v5 is out of alpha if you wish :)

Hyperkid123:
> I'm also not sure how you want to handle this. Since MUI 5 has a dep on react 17, I'm not sure you want to force consumers of this to upgrade all their projects to 17. so maybe a separate mui-17-component-mapper package?

@guyathomas react@17 "should" be backwards compatible with react@16 so I think using peer dependency >=16.8.x will be good enough for now.

We want to update to lts version of node soon. The reason why its locked is because of firebase hosting for the docs packages. We didn't have a chance to id yet.

> Still very much WIP, and happy to close until v5 is out of alpha if you wish :)

Yeah, until there is a stable version of the MUI library we won't be releasing a stable version of the mapper. But we can have a pre-release version, I see no problem with it. I will create an MUI-5 branch for you.

thecodejack:
>@guyathomas react@17 "should" be backwards compatible with react@16 so I think using peer dependency >=16.8.x will be good enough for now.

Actually React 17 is minimum dependancy for material ui v5. So atleast material ui wrapper module may need to end up as major release to avoid this breaking change. Probably node js support also can be updated along same.

Issue #1038, opened 2021-04-19
Stackblitz examples with react-jss do not work


rvsia:
**Scope**: demo

**Description**

![Screenshot 2021-04-16 at 16 01 04](https://user-images.githubusercontent.com/32869456/115036494-de4cee00-9ecd-11eb-8db6-eab96a3269ea.png)

I contacted Stackblitz and they will investigate.

rvsia:
similar issue: https://github.com/cssinjs/jss/issues/1329

PR #1083, opened 2021-06-14
fix(common): transform react-jss to esm/cjs

rvsia:
Fixes #https://github.com/data-driven-forms/react-forms/issues/1038

**Description**

- this should fix the errors in the example


Hyperkid123 (approved):
:crossed_fingers:

PR #1093, opened 2021-06-23
Fix wrong space character in demo

rvsia:
My Mac loves to produce this stupid space that is not a space.


Hyperkid123 (approved):
:space_invader:

PR #1101, opened 2021-06-28
Store selected schema in query, not in state

rvsia:
**Description**

Stores the active schema in query so people can share links


rvsia:
This PR ensures that these variants will be excluded from the search results (no duplicates): https://github.com/algolia/docsearch-configs/pull/4276

Hyperkid123 (approved):

PR #1110, opened 2021-08-20
Fix wrong set of Date

### LINKED ISSUE #1109: Form Renderer - Custom Implementation - Date Picker and Date object Issue

Form renderer

**Description**

I am having a custom implementation of Form Renderer. I bumped into a weird issue related to the Date Picker and Date object. Hope someone here can shed some light on what is causing the issue. 

Basically, the Date Picker is expecting or outputting a Date object as value. 

If initialValue is NOT provided, everything works fine and user can set the value or change the value using the Date Picker. If initialValue is provided and users do not touch the input, it is also ok. However, if users touch the input, the underlying value becomes **undefined**.  

An example is available at:
https://codesandbox.io/s/eager-meadow-qli49?file=/src/DateComponent.tsx:227-684

Screen Shot 2021-08-19 at 9 19 11 PM

I have added some debug information in the source. I can confirm that the DatePicker onChange event is triggered in all the cases. But the value just disappear when the initialvalue is provided and users change the input. 

This is the output of getState().values when the form is rendered. 

Screen Shot 2021-08-19 at 9 32 14 PM

This is the output of getState().values when users select a date from DatePicker. 

Screen Shot 2021-08-19 at 9 32 27 PM

This is the output in the onChange handler of the DatePicker. 

Screen Shot 2021-08-19 at 9 34 49 PM

**Schema**

const schema = {
  fields: [
    {
      component: componentTypes.DATE_PICKER,
      name: "date1",
      label: "Date 1",
      validate: [
        {
          type: validatorTypes.REQUIRED
        }
      ]
    },
    {
      component: componentTypes.DATE_PICKER,
      name: "date2",
      label: "Date 2",
      validate: [
        {
          type: validatorTypes.REQUIRED
        }
      ]
    }
  ]
};

const initialValue = {
  date1: new Date()
};

<If it's possible, please provide a schema which reproduces the issue>

mmarkelov:
Fix #1109 

I found out that there is unexpected behavior for passed date objects. It will be handled as empty object in `checkEmpty` function. These is because of `isEmpty` function from **lodash** - it will return **true** for Date. https://github.com/lodash/lodash/issues/5131

### DIFF
--- a/packages/react-form-renderer/src/tests/form-renderer/enhanced-on-change.test.js
+++ b/packages/react-form-renderer/src/tests/form-renderer/enhanced-on-change.test.js
@@ -29,7 +29,7 @@ describe('#enhancedOnChange', () => {
     expect(enhancedOnChange({ onChange: (value) => value, clearedValue }, value)).toEqual('Me');
   });
 
-  it('should return booelan events correctly with initialValue set', () => {
+  it('should return boolean events correctly with initialValue set', () => {
     const initial = false;
     const valueFalse = {
       target: {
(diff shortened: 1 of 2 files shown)


rvsia:
Hello @mmarkelov! Thank you so much for this PR! ❤️ 

Can you please change the commit message to `fix(renderer): Fix wrong set of Date` so a new build will be triggered?

rvsia (approved):
Looks great! Thank you! 🏆

PR #1114, opened 2021-08-30
Add description to pf select

### LINKED ISSUE #1113: Add "Description" to PF4 Select
PF4 mapper

**Description**
There is a need to add the Description field to the PF4 Select structure in DDF as this should give us the functionality of the Patternfly Drop down component with descriptions 
https://www.patternfly.org/v4/components/dropdown#dropdown-with-descriptions

Screen Shot 2021-08-30 at 11 11 42 AM

**Schema**

<If it's possible, please provide a schema which reproduces the issue>

rvsia:
Fixes #1113 

**Description**

Adds description to PF4 select options.

**Schema** *(if applicable)*

```jsx
options: [
  { value: 'value', label: 'Some label', description: 'Some description' }
]
```

Screenshot 2021-08-30 at 19 08 45

### DIFF
--- a/packages/pf4-component-mapper/src/select/select/option.js
+++ b/packages/pf4-component-mapper/src/select/select/option.js
@@ -1,8 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import './option.css';
-
 import { CheckIcon } from '@patternfly/react-icons';
 
 const Option = ({ item, isActive, isSelected, ...props }) => (
@@ -21,7 +19,7 @@ const Option = ({ item, isActive, isSelected, ...props }) => (
           <CheckIcon />
         </span>
       )}
-      {item.description && <div className="ddorg__pf4-component-mapper__select-option-description">{item.description}</div>}
+      {item.description && <div className="pf-c-select__menu-item-description">{item.description}</div>}
     </button>
   </li>
 );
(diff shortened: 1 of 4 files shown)


packages/pf4-component-mapper/src/select/select/option.js
```diff
@@ -19,6 +21,7 @@ const Option = ({ item, isActive, isSelected, ...props }) => (
           <CheckIcon />
         </span>
       )}
+      {item.description && <div className="ddorg__pf4-component-mapper__select-option-description">{item.description}</div>}
```
Hyperkid123, line 24: @rvsia is it possible to use the PF class for this? As we do with the rest? I think we will hear about this not being a PF component if we don't use them.

rvsia: Yeah, that was the first thing I tried to do. However, we do not import the dropdown component so the PF class styles are missing. (And the whole select is custom styled anyways..)

rvsia: *the class was there, but the dropdown specific variables were missing.

Hyperkid123: Yup, we should use the class. And you can always import the CSS from @patternfly/react-styles to inject them.

Hyperkid123 (approved):

PR #1125, opened 2021-09-15
Update to next11

rvsia:


Hyperkid123 (approved):

packages/react-renderer-demo/src/next.config.js
```diff
@@ -69,7 +69,6 @@ module.exports = withBundleAnalyzer(
 
         config.optimization.minimizer = [
           new TerserPlugin({
-            cache: true,
```
rvsia, line 72: @Hyperkid123 was this important? 😄 The option does not exist anymore and there is no migration guide to it.

Hyperkid123: NO, it was not. There is the persistent environment (it's not configured) so there is nothing to cache really.

PR #1137, opened 2021-10-11
feat(carbon): add string variant of timepicker

rvsia:
Fixes **see discord**

**Description**

### useStringFormat

If set to **true**, then the value is stored as a **string**. You need to parse it yourself. You should also provide a proper validation, currently the component does not check what users enter.

Examples:

`12:04`

`12:04 PM`

`12:04 PM EST` (timezone's value is being used)

<br />

If not set (or set to **false**), then the value is stored as a **Date**. You need to use `getHours` (see [more](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getHours)) and `getMinutes` (see [more](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMinutes)) to obtain the values in your submit function.

**Schema** *(if applicable)*

```jsx
            schema={{
              fields: [
                {
                  component: 'time-picker',
                  name: 'time-picker',
                  twelveHoursFormat: true,
                  timezones: [
                    { label: 'EST', value: 'est' },
                    { label: 'PST', value: 'pst' },
                  ],
                  useStringFormat: true,
                },
              ],
            }}
```

### DIFF
--- a/packages/carbon-component-mapper/src/tests/time-picker-string.test.js
+++ b/packages/carbon-component-mapper/src/tests/time-picker-string.test.js
@@ -1,0 +1,241 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import { FormRenderer, componentTypes } from '@data-driven-forms/react-form-renderer';
+
+import FormTemplate from '../form-template';
+import { act } from 'react-dom/test-utils';
+import TimePicker from '../time-picker';
+
+describe('TimePicker<String>', () => {
+  let initialProps;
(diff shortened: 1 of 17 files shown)


GilbertCherrie:
@rvsia This pr also needs to be back-ported to branch v2 should I include it in my pr?

rvsia:
@GilbertCherrie yep, include anything you need

PR #1144, opened 2021-10-13
Fix hidden label in PF4 field array

### LINKED ISSUE #1141: PF4-Mapper: conditions within a field-array do not hide the label
**Scope**: PF4 mapper

**Description**
When using a condition to hide an input from within a form-array the label will remain while the input get's hidden. This is due to the fact that the label is handled outside of the rendering of the input. I'm placing this here in case others might have an approach to resolving this (I can't see a clear path).

The only approach I can see would be to do an application of the conditions before a render call, perhaps something like a `preRender()`.

**Schema**
```
const schema = {
  fields: [
    {
      "component": "field-array",
      "label": "Select One",
      "name": "SelectOne",
      "fields": [
        {
          "component": "select",
          "label": "Select Option",
          "name": "Option",
          "simpleValue": true,
          "options": [
            {
              "label": "Placeholder1",
              "value": "Placeholder1"
            },
            {
              "label": "Placeholder2",
              "value": "Placeholder2"
            },
          ]
        },
        {
          "component": "text-field",
          "name": "HiddenInput",
          "label": "Hidden Input",
          "condition": {
            when: `Option`,
            is: 'Placeholder1',
            then: {visible: true},
          }
        },
      ]
    } 
  ]
}
```

[Sandbox Example](https://codesandbox.io/s/poc-forked-40bdj?file=/src/App.js)

rvsia:
Fixes #1141 

**Description**

`FieldArray` items are renderer in special wrapper (`FieldArrayItem`) that uses the field condition.

**After** (the value stays because of no `clearOnUnmount`)

Kapture 2021-10-13 at 11 42 02

**Schema** *(if applicable)*

```jsx
schema: {
        fields: [
          {
            component: componentTypes.FIELD_ARRAY,
            name: 'hide-field',
            fields: [
              {
                name: 'select',
                label: 'Shown label',
              },
              {
                name: 'hidden',
                label: 'Hidden label',
                condition: {
                  when: ({ name }) => name.replace('hidden', 'select'),
                  isNotEmpty: true,
                },
              },
            ],
          },
        ],
      },
```

### DIFF
--- a/packages/react-renderer-demo/src/pages/schema/constants.md
+++ b/packages/react-renderer-demo/src/pages/schema/constants.md
@@ -20,6 +20,7 @@ Available keys:
 {
   TEXT_FIELD,
   FIELD_ARRAY,
+  FIELD_ARRAY_ITEM,
   CHECKBOX,
   SUB_FORM,
   RADIO,


rvsia:
@Hyperkid123 Actually, PF has implemented its own design and we should probably adapt it - and in that case, this would be no longer an issue. 😱 

https://www.patternfly.org/v4/components/form#field-groups

PR #1150, opened 2021-10-15
fix(carbon): fix time picker minute range

GilbertCherrie:
**Description**

Changed minute overflow calculation to occur at 60 minutes instead of 59.


GilbertCherrie:
@Hyperkid123 Is there any chance this PR can be backported to version 2 of DDF. Only asking because we need this for older versions of ManageIQ currently using DDF version 2.

PR #1160, opened 2021-10-26
fix(carbon): fix initial value for date timepicker

### LINKED ISSUE #1155: Carbon Component Mapper Time Picker Initial Value Bug
Scope: Carbon Component Mapper

**Description**
For setting an initial value for the time picker component, when setting any times that are PM the time picker will set the correct time (05:08 for example) but the AM/PM dropdown will not change to PM, it is always AM no matter the initial time. I have an example here:
https://stackblitz.com/edit/x6mwm1-ha1ag5?file=index.js (edited)

rvsia:
Fixes #1155

**Schema** *(if applicable)*

```jsx
const date = new Date();
date.setHours(18);
const date2 = new Date();
date2.setHours(1);

            schema={{
              fields: [
                { name: 'time-picker-1', component: componentTypes.TIME_PICKER, twelveHoursFormat: true, initialValue: date },
                { name: 'time-picker-2', component: componentTypes.TIME_PICKER, twelveHoursFormat: true, initialValue: date2 },
              ],
            }}
```

### DIFF
--- a/packages/carbon-component-mapper/src/tests/time-picker-string.test.js
+++ b/packages/carbon-component-mapper/src/tests/time-picker-string.test.js
@@ -193,7 +193,7 @@ describe('TimePicker<String>', () => {
     onSubmit.mockReset();
 
     await act(async () => {
-      wrapper.find('select#time-picker-12h').simulate('change', { target: { value: 'UTC' } });
+      wrapper.find('select#time-picker-timezones').simulate('change', { target: { value: 'UTC' } });
     });
     wrapper.update();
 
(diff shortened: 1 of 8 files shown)


GilbertCherrie:
@Hyperkid123 Is there any chance this PR can be backported to version 2 of DDF. Only asking because we need this for older versions of ManageIQ currently using DDF version 2.

rvsia:
@GilbertCherrie can you please prepare a PR with both fixes? Against branch `v2`, I can release it tomorrow.

PR #1166, opened 2021-11-03
Update types generator to use correct variable/function names.

### LINKED ISSUE #1165: Relative imports can cause issues during builds.
### Discussed in https://github.com/data-driven-forms/react-forms/discussions/1164

<div type='discussions-op-text'>

<sup>Originally posted by **steverhoades** October 31, 2021</sup>
I'm curious if it was a design choice to not have these available from the base package.
e.g. I must explicitly import this from `@data-driven-forms/react-form-renderer/use-form-api` or  `@data-driven-forms/react-form-renderer/use-field-api` respectively vs. `import {useFieldAPi} from @data-driven-forms/react-form-renderer`.

I only ask as I've spent the last hour or so debugging why my microapp wasn't playing nice with it's application shell. Turns out it was moving the dependencies from the microapp to the shell that broke as I didn't explicitly declare `use-form-api` and `use-field-api` as dependency externals.</div>

Hyperkid123:
Fixes #1165

**Description**
Functions and constants had incorrect type name casing resulting in missing TS types when using named exports. For example:
```TS
// results in TS error that the module does not exist
import { useFieldApi } from '@data-driven-forms/react-form-renderer'
// the type was named but that function does not exist
import { UseFieldApi } from '@data-driven-forms/react-form-renderer'
```
### Changes
- the default re-export name is based on the actual variable/function name instead of automatically converting everything to CamelCase. If the script is unable to extract the correct name, it will fall back to CamelCase conversion.
- fixed some invalid type definitions

- [x] `Yarn build` passes
- [x] `Yarn lint` passes
- [x] `Yarn test` passes
- [x] Correct commit message
   - scope: package

cc @steverhoades

### DIFF
--- a/packages/ant-component-mapper/src/validation-error/validation-error.d.ts
+++ b/packages/ant-component-mapper/src/validation-error/validation-error.d.ts
@@ -1,5 +1,5 @@
 import { Meta } from "@data-driven-forms/react-form-renderer";
 
-export type validationError = (meta: Meta<any>, validateOnMount?: boolean) => boolean | any | undefined
+export function validationError(meta: Meta<any>, validateOnMount?: boolean): boolean | any | undefined
 
 export default validationError;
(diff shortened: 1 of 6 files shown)


rvsia (approved):
🏆

PR #1170, opened 2021-11-05
fix(pf4): use options from field hook

rvsia:
Options were taken from props, not from enhanced props, so `resolveProps` didn't work here.


karelhala (approved):
:green_circle:

PR #1183, opened 2022-01-06
fix(common): update common version in package.json

rvsia:
We can replace common version in package.json using [`semantic-release/exec`](https://github.com/semantic-release/exec/tree/master/test) plugin. This will prevent issues with mismatched package versions as seen in Edge frontend.

This plugin takes the `nextVersion`, finds all `package.json`s in `packages/package/` folders and changes the `*` to the `nextVersion`.

I tested the script directly and it *worked*, but I wasn't able to test it in dry-run, I guess we have to just hope. @Hyperkid123 Please do not merge it when we are not online so if there is any issue, we can deprecate the wrong versions. 😄 

This step happens **after git commit** so the changes should not be committed to the repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "@khala/commit-analyzer-wildcard": "^2.5.2",
     "@khala/npm-release-monorepo": "^2.5.2",
     "@khala/wildcard-release-notes": "^2.5.2",
+    "@semantic-release/exec": "^6.0.3",
     "@semantic-release/git": "^9.0.1",
     "@semantic-release/github": "^7.2.3",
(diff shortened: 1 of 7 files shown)


Hyperkid123:
@rvsia the glob pattern looks good I think we can safely merge this. Though I think we should do the same for the renderer package as every other package depends on it.

rvsia:
@Hyperkid123 renderer is a peer dependency

Hyperkid123:
@rvsia yes but npm@7 tries to install peer dependencies as well which may cause some unwanted installation warnings/errors.

Hyperkid123 (approved):

PR #1218, opened 2022-03-11
feat(react): pass props and child render props

zgover:
# Description

- fix: respect of children prop to support subscribing to contexts through child components rather than only `FormTemplate` or fields.
- feat: optimized portability of `FormRender` by supporting `FormTemplate` pass prop through `FormTemplateProps`.
- feat: overridable form rendering by passing children as a render function.

Please include a summary of the change.

**Schema** *(if applicable)*

```jsx
// Children prop as ReactElement
const MyOtherChild = ({allowReset}) => {
    const {onReset} = useFormApi()
   return !allowReset ? null : (
       <button onClick={() => onReset()}/>Reset</button>
   )
}
<FormRenderer
    FormTemplate={AuthFormTemplateComponent}
    componentMapper={simpleComponentMapper}
    onSubmit={handleFormSubmit}
    schema={formSchema}
    subscription={{values: true}}
    clearOnUnmount
>
    <MyOtherChild allowReset/>
</FormRenderer>
```

```jsx
// Template pass props
<FormRenderer
    FormTemplate={AuthFormTemplateComponent}
    FormTemplateProps={{type: 'signup'}} // signout | signup => form variant
    componentMapper={simpleComponentMapper}
    onSubmit={handleFormSubmit}
    schema={formSchema}
    subscription={{values: true}}
    clearOnUnmount
  />
```

```jsx
// Children render prop
<FormRenderer
    componentMapper={simpleComponentMapper}
    onSubmit={handleFormSubmit}
    schema={formSchema}
    subscription={{values: true}}
    clearOnUnmount
>
    {({formFields, schema}) => (
        <AuthFormTemplateComponent
            schema={schema}
            formFields={formFields}
            type={'signup'}
       />
    )}
</FormRenderer>
```

- [ ] `Yarn build` passes
- [ ] `Yarn lint` passes
- [ ] `Yarn test` passes
- [ ] Test coverage for new code *(if applicable)*
- [ ] Documentation update *(if applicable)*
- [ ] Correct commit message
   - scope: package

### DIFF
--- a/packages/react-form-renderer/src/form-renderer/form-renderer.js
+++ b/packages/react-form-renderer/src/form-renderer/form-renderer.js
@@ -96,7 +96,7 @@ const FormRenderer = ({
 
   const internalGetRegisteredFields = useCallback(() => {
     const fields = registeredFields.current;
-    Object.entries(fields).reduce((acc, [name, value]) => (
+    return Object.entries(fields).reduce((acc, [name, value]) => (
       value > 0 ? [...acc, name] : acc
     ), []);
   }, []);
--- a/packages/react-form-renderer/src/form-renderer/form-renderer.js
+++ b/packages/react-form-renderer/src/form-renderer/form-renderer.js
@@ -50,7 +50,7 @@ const FormRenderer = ({
   ]), [decorators]);
 
   const handleSubmitCallback = useCallback((values, formApi, ...args) => {
-    return !isFunc(onSubmit) ? void 0 : onSubmit(values, {...formApi, fileInputs}, ...args);
+    return !isFunc(onSubmit) ? undefined : onSubmit(values, {...formApi, fileInputs}, ...args);
   }, [onSubmit, fileInputs]);
 
   const handleCancelCallback = useCallback((getState) => (...args) => {
@@ -86,13 +86,13 @@ const FormRenderer = ({
     setRegisteredFields((prev) => (
       prev[name] ? {...prev, [name]: prev[name] + 1} : {...prev, [name]: 1})
     );
-  }, [setRegisteredFields]);
+  }, []);
 
   const internalUnRegisterField = useCallback((name) => {
     setRegisteredFields(({[name]: currentField, ...prev}) => (
       currentField && currentField > 1 ? {[name]: currentField - 1, ...prev} : prev
     ));
-  }, [setRegisteredFields]);
+  }, []);
 
   const internalGetRegisteredFields = useCallback(() => {
     const fields = registeredFields.current;
--- a/packages/react-form-renderer/src/form-renderer/form-renderer.js
+++ b/packages/react-form-renderer/src/form-renderer/form-renderer.js
@@ -1,79 +1,132 @@
-import React, { useState, useRef } from 'react';
-import Form from '../form';
 import arrayMutators from 'final-form-arrays';
-import PropTypes from 'prop-types';
 import createFocusDecorator from 'final-form-focus';
+import PropTypes from 'prop-types';
+import React, {useCallback, useMemo, useRef, useState} from 'react';
 
+import defaultSchemaValidator from '../default-schema-validator';
+import defaultValidatorMapper from '../validator-mapper';
+import Form from '../form';
 import RendererContext from '../renderer-context';
 import renderForm from './render-form';
-import defaultSchemaValidator from '../default-schema-validator';
 import SchemaErrorComponent from './schema-error-component';
-import defaultValidatorMapper from '../validator-mapper';
+
+
+const isFunc = (fn) => (typeof fn === 'function');
 
 const FormRenderer = ({
+  actionMapper,
   children,
+  clearedValue,
+  clearOnUnmount,
   componentMapper,
+  decorators,
   FormTemplate,
   FormTemplateProps,
-  onSubmit,
+  mutators,
   onCancel,
+  onError,
   onReset,
-  clearOnUnmount,
-  subscription,
-  clearedValue,
+  onSubmit,
   schema,
-  validatorMapper,
-  actionMapper,
   schemaValidatorMapper,
+  subscription,
+  validatorMapper,
   ...props
 }) => {
   const [fileInputs, setFileInputs] = useState([]);
   const registeredFields = useRef({});
   const focusDecorator = useRef(createFocusDecorator());
-  let schemaError;
+  const validatorMapperMerged = useMemo(() => ({
+    ...defaultValidatorMapper,
+    ...validatorMapper
+  }), [validatorMapper]);
+  const mutatorsMerged = useMemo(() => ({
+    ...arrayMutators,
+    ...mutators
+  }), [mutators]);
+  const decoratorsMerged = useMemo(() => ([
+    focusDecorator.current,
+    ...(Array.isArray(decorators) ? decorators : [])
+  ]), [decorators]);
 
-  const setRegisteredFields = (fn) => (registeredFields.current = fn({ ...registeredFields.current }));
-  const internalRegisterField = (name) => {
-    setRegisteredFields((prev) => (prev[name] ? { ...prev, [name]: prev[name] + 1 } : { ...prev, [name]: 1 }));
-  };
+  const handleSubmitCallback = useCallback((values, formApi, ...args) => {
+    return !isFunc(onSubmit) ? void 0 : onSubmit(values, {...formApi, fileInputs}, ...args);
+  }, [onSubmit, fileInputs]);
 
-  const internalUnRegisterField = (name) => {
-    setRegisteredFields(({ [name]: currentField, ...prev }) => (currentField && currentField > 1 ? { [name]: currentField - 1, ...prev } : prev));
-  };
+  const handleCancelCallback = useCallback((getState) => (...args) => {
+    return !isFunc(onCancel) ? void 0 : onCancel(getState().values, ...args);
+  }, [onCancel]);
 
-  const internalGetRegisteredFields = () =>
-    Object.entries(registeredFields.current).reduce((acc, [name, value]) => (value > 0 ? [...acc, name] : acc), []);
+  const handleResetCallback = useCallback((reset) => (...args) => {
+    reset();
+    return !isFunc(onReset) ? void 0 : onReset(...args);
+  }, [onReset]);
 
-  const validatorMapperMerged = { ...defaultValidatorMapper, ...validatorMapper };
+  const handleErrorCallback = useCallback((...args) => {
+    // eslint-disable-next-line no-console
+    console.error(...args);
+    return !isFunc(onError) ? void 0 : onError(...args);
+  }, [onError]);
+
+  const registerInputFile = useCallback((name) => {
+    setFileInputs((prevFiles) => [...prevFiles, name]);
+  }, []);
+
+  const unRegisterInputFile = useCallback((name) => {
+    setFileInputs((prevFiles) => [
+      ...prevFiles.splice(prevFiles.indexOf(name))
+    ]);
+  }, []);
+
+  const setRegisteredFields = useCallback((fn) => {
+    return registeredFields.current = fn({...registeredFields.current});
+  }, []);
+
+  const internalRegisterField = useCallback((name) => {
+    setRegisteredFields((prev) => (
+      prev[name] ? {...prev, [name]: prev[name] + 1} : {...prev, [name]: 1})
+    );
+  }, [setRegisteredFields]);
+
+  const internalUnRegisterField = useCallback((name) => {
+    setRegisteredFields(({[name]: currentField, ...prev}) => (
+      currentField && currentField > 1 ? {[name]: currentField - 1, ...prev} : prev
+    ));
+  }, [setRegisteredFields]);
+
+  const internalGetRegisteredFields = useCallback(() => {
+    const fields = registeredFields.current;
+    Object.entries(fields).reduce((acc, [name, value]) => (
+      value > 0 ? [...acc, name] : acc
+    ), []);
+  }, []);
 
   try {
     const validatorTypes = Object.keys(validatorMapperMerged);
     const actionTypes = actionMapper ? Object.keys(actionMapper) : [];
-    defaultSchemaValidator(schema, componentMapper, validatorTypes, actionTypes, schemaValidatorMapper);
-  } catch (error) {
-    schemaError = error;
-    // eslint-disable-next-line no-console
-    console.error(error);
-    // eslint-disable-next-line no-console
-    console.log('error: ', error.message);
+
+    defaultSchemaValidator(
+      schema,
+      componentMapper,
+      validatorTypes,
+      actionTypes,
+      schemaValidatorMapper
+    );
+  }
+  catch (error) {
+    handleErrorCallback('schema-error', error);
+    return <SchemaErrorComponent name={error.name} message={error.message} />;
   }
 
-  if (schemaError) {
-    return <SchemaErrorComponent name={schemaError.name} message={schemaError.message} />;
-  }
-
-  const registerInputFile = (name) => setFileInputs((prevFiles) => [...prevFiles, name]);
-
-  const unRegisterInputFile = (name) => setFileInputs((prevFiles) => [...prevFiles.splice(prevFiles.indexOf(name))]);
+  const formFields = useMemo(() => renderForm(schema.fields), [schema]);
 
   return (
     <Form
-      {...props}
-      onSubmit={(values, formApi, ...args) => onSubmit(values, { ...formApi, fileInputs }, ...args)}
-      mutators={{ ...arrayMutators }}
-      decorators={[focusDecorator.current]}
-      subscription={{ pristine: true, submitting: true, valid: true, ...subscription }}
-      render={({ handleSubmit, pristine, valid, form: { reset, mutators, getState, submit, ...form } }) => (
+      onSubmit={handleSubmitCallback}
+      mutators={mutatorsMerged}
+      decorators={decoratorsMerged}
+      subscription={{pristine: true, submitting: true, valid: true, ...subscription}}
+      render={({handleSubmit, pristine, valid, form: {reset, mutators, getState, submit, ...form}}) => (
         <RendererContext.Provider
           value={{
             componentMapper,
@@ -84,11 +137,9 @@ const FormRenderer = ({
               unRegisterInputFile,
               pristine,
               onSubmit,
-              onCancel: onCancel ? (...args) => onCancel(getState().values, ...args) : undefined,
-              onReset: (...args) => {
-                onReset && onReset(...args);
-                reset();
-              },
+              onCancel: handleCancelCallback(getState),
+              onReset: handleResetCallback(reset),
+              onError: handleErrorCallback,
               getState,
               valid,
               clearedValue,
@@ -107,33 +158,41 @@ const FormRenderer = ({
             },
           }}
         >
-          {typeof children === 'function' ? children({schema, formFields: renderForm(schema.fields)}) : (
-            <React.Fragment>
-              <FormTemplate
-                formFields={renderForm(schema.fields)}
-                schema={schema}
-                {...FormTemplateProps}
-              />
-              {children}
-            </React.Fragment>
+
+          {FormTemplate && (
+            <FormTemplate
+              formFields={formFields}
+              schema={schema}
+              {...FormTemplateProps}
+            />
           )}
+
+          {isFunc(children) ? children({formFields, schema}) : children}
+
         </RendererContext.Provider>
       )}
+      {...props}
     />
   );
 };
 
 FormRenderer.propTypes = {
   children: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
-  onSubmit: PropTypes.func.isRequired,
+  onSubmit: PropTypes.func,
   onCancel: PropTypes.func,
   onReset: PropTypes.func,
+  onError: PropTypes.func,
   schema: PropTypes.object.isRequired,
   clearOnUnmount: PropTypes.bool,
-  subscription: PropTypes.shape({ [PropTypes.string]: PropTypes.bool }),
+  subscription: PropTypes.shape({[PropTypes.string]: PropTypes.bool}),
   clearedValue: PropTypes.any,
   componentMapper: PropTypes.shape({
-    [PropTypes.string]: PropTypes.oneOfType([PropTypes.node, PropTypes.element, PropTypes.func, PropTypes.elementType]),
+    [PropTypes.string]: PropTypes.oneOfType([
+      PropTypes.node,
+      PropTypes.element,
+      PropTypes.func,
+      PropTypes.elementType
+    ]),
   }).isRequired,
   FormTemplate: PropTypes.elementType,
   FormTemplateProps: PropTypes.object,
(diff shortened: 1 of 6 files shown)


zgover:
@Hyperkid123 the linter is way too strict on whitespace, it does not want to allow a multiline ternary boolean. Personally, I find it very disorganized whenever a lines span 100 while using xml syntax. How do you want to proceed with this?

Hyperkid123 (commented):
@zgover thanks for the PR. I am all for this feature however we need a few changes.

1. By removing the required prop from the form template, we now have to check if the `FormTemplate` or `children` prop exists. You can probably add a runtime check as we do with the schema validator on the top of the component.
2. Add a unit test
3. We have to add documentation for this change
4. The linter issue is not about maximum line length but because the "else" branch of the ternary operator is a multiline expression (JSX). I think the proposed fix is not that bad.

zgover:
@Hyperkid123 before I proceed—see last commit; I want to make sure you're ok with these changes.

Hyperkid123:
@zgover I am more than OK with the last commit. these changes are something we should have done a long time ago.

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -1,77 +1,132 @@
...
+    return registeredFields.current = fn({...registeredFields.current});
+  }, []);
 
-  const internalGetRegisteredFields = () =>
-    Object.entries(registeredFields.current).reduce((acc, [name, value]) => (value > 0 ? [...acc, name] : acc), []);
+  const internalRegisterField = useCallback((name) => {
+    setRegisteredFields((prev) => (
+      prev[name] ? {...prev, [name]: prev[name] + 1} : {...prev, [name]: 1})
+    );
+  }, [setRegisteredFields]);
```
rvsia, line 89: `setRegisteredFields` cannot be changed, so empty array is enough here

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -1,77 +1,132 @@
...
+    ...arrayMutators,
+    ...mutators
+  }), [mutators]);
+  const decoratorsMerged = useMemo(() => ([
+    focusDecorator.current,
+    ...(Array.isArray(decorators) ? decorators : [])
+  ]), [decorators]);
+
+  const handleSubmitCallback = useCallback((values, formApi, ...args) => {
+    return !isFunc(onSubmit) ? void 0 : onSubmit(values, {...formApi, fileInputs}, ...args);
```
rvsia, line 53: personally I would return `undefined` instead of `void 0`, we do not use it anywhere else in the codebase

rvsia, line 53: personally I would return `undefined` instead of `void 0`, we do not use it anywhere else in the codebase

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -1,77 +1,132 @@
...
+      prev[name] ? {...prev, [name]: prev[name] + 1} : {...prev, [name]: 1})
+    );
+  }, [setRegisteredFields]);
 
-  const validatorMapperMerged = { ...defaultValidatorMapper, ...validatorMapper };
+  const internalUnRegisterField = useCallback((name) => {
+    setRegisteredFields(({[name]: currentField, ...prev}) => (
+      currentField && currentField > 1 ? {[name]: currentField - 1, ...prev} : prev
+    ));
+  }, [setRegisteredFields]);
```
rvsia, line 95: ^^^

rvsia, line 95: ^^^

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -1,77 +1,132 @@
...
   ...props
 }) => {
   const [fileInputs, setFileInputs] = useState([]);
   const registeredFields = useRef({});
   const focusDecorator = useRef(createFocusDecorator());
-  let schemaError;
+  const validatorMapperMerged = useMemo(() => ({
+    ...defaultValidatorMapper,
+    ...validatorMapper
+  }), [validatorMapper]);
```
rvsia, line 42: Not sure if dependency array will work well with an object, the same applies for mutators/decorators (arrays)

rvsia, line 42: Not sure if dependency array will work well with an object, the same applies for mutators/decorators (arrays)

packages/react-form-renderer/src/form-renderer/form-renderer.js
zgover: Updated in last commit

packages/react-form-renderer/src/form-renderer/form-renderer.js
zgover: Updated in last commit

packages/react-form-renderer/src/form-renderer/form-renderer.js
zgover: Updated in last commit

packages/react-form-renderer/src/form-renderer/form-renderer.js
zgover: Elaborate?

zgover:
Last commit uncluded @rvsia suggestions, haven't had a chance to really look over the build error logs yet but just at the short glance looks like it may just be tests that need to be updated or a fix is due in one of my updates.

packages/react-form-renderer/src/form-renderer/form-renderer.js
rvsia: The `useMemo` will run comparison between two arrays/objects. In most cases these arrays/objects will be recreated as a unnamed prop on every render, so they will be different even nothing changes. But it's not a big deal.

zgover:
@Hyperkid123 @rvsia, I'm not too familiar with the current tests that are set up and entire submit flow. So it may be easier for y'all to determine the major issues here in the failed build.

For example it appears the submit handler is being called but there seems to be an additional field value of `undefined` in the values being passed to all of them:
```
 FAIL  packages/suir-component-mapper/src/tests/wizard.test.js 
  ● wizard › conditional submit

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    - Expected
    + Received

      Object {
    -   "aws": "aws",
    -   "summary": "summary",
    +   "undefined": undefined,
      },

...
```

Secondly, it appears the only validation for the schema error render is checking the console logs:
```
FAIL  packages/react-form-renderer/src/tests/form-renderer/form-renderer.test.js
  ● <FormRenderer /> › should render errorComponent form from schema

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    Expected: "error: ", Any<String>

    Number of calls: 0

      105 |     expect(screen.getByText('Form could not be rendered, because of invalid form schema.')).toBeInTheDocument();
      106 |     expect(spy).toHaveBeenCalled();
    > 107 |     expect(logSpy).toHaveBeenCalledWith('error: ', expect.any(String));
          |                    ^
      108 |
      109 |     console = _console; // eslint-disable-line
      110 |   });

      at Object.<anonymous> (packages/react-form-renderer/src/tests/form-renderer/form-renderer.test.js:107:20)
```

Hyperkid123:
@zgover 

> For example it appears the submit handler is being called but there seems to be an additional field value of undefined in the values being passed to all of them:

```
 FAIL  packages/suir-component-mapper/src/tests/wizard.test.js 
  ● wizard › conditional submit
```

So in this test, the submit callback does not receive correct values. Not only there is an extra `undefined` but the values that should have been in the values object are entirely missing. So the submit call back is broken

I will checkout the PR and see why the submit function has invalid argument

> Secondly, it appears the only validation for the schema error render is checking the console logs:

That is not the case. The expect a couple of lines above also checks if the error component was rendered.
```
expect(screen.getByText('Form could not be rendered, because of invalid form schema.')).toBeInTheDocument();
```

Hyperkid123:
@zgover the issue with the wizard test is in the `internalGetRegisteredFields` function. Let me check the changes and why is not doing what it's supposed to.

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -1,77 +1,132 @@
...
-  const validatorMapperMerged = { ...defaultValidatorMapper, ...validatorMapper };
+  const internalUnRegisterField = useCallback((name) => {
+    setRegisteredFields(({[name]: currentField, ...prev}) => (
+      currentField && currentField > 1 ? {[name]: currentField - 1, ...prev} : prev
+    ));
+  }, []);
+
+  const internalGetRegisteredFields = useCallback(() => {
+    const fields = registeredFields.current;
+    Object.entries(fields).reduce((acc, [name, value]) => (
```
Hyperkid123, line 99: @zgover  This function is missing `return` that is why the submit tests in wizard are failing.

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -1,77 +1,132 @@
...
+    return !isFunc(onReset) ? void 0 : onReset(...args);
+  }, [onReset]);
 
-  const setRegisteredFields = (fn) => (registeredFields.current = fn({ ...registeredFields.current }));
-  const internalRegisterField = (name) => {
-    setRegisteredFields((prev) => (prev[name] ? { ...prev, [name]: prev[name] + 1 } : { ...prev, [name]: 1 }));
-  };
+  const handleErrorCallback = useCallback((...args) => {
+    // eslint-disable-next-line no-console
+    console.error(...args);
```
Hyperkid123, line 67: The original code is calling the `console.log` with just the error message without the whole stack
```js
  try {
    const validatorTypes = Object.keys(validatorMapperMerged);
    const actionTypes = actionMapper ? Object.keys(actionMapper) : [];
    defaultSchemaValidator(schema, componentMapper, validatorTypes, actionTypes, schemaValidatorMapper);
  } catch (error) {
    schemaError = error;
    // eslint-disable-next-line no-console
    console.error(error);
    // eslint-disable-next-line no-console
    console.log('error: ', error.message);
  }
  ```
That is why the second test is falling. I am honestly OK with just using the `console.error`. @rvsia do you have any objections to removing the `console.log` part? If no we can just drop the expect from the test

Comment (zgover) [reply packages/react-form-renderer/src/form-renderer/form-renderer.js:67]: @rvsia ?

Comment (Hyperkid123) [reply packages/react-form-renderer/src/form-renderer/form-renderer.js:67]: @zgover sorry I was on vacation. Just delete the console log check from tests. It's not necessary to log the same message multiple times.

**Comment by zgover** [comment]
@Hyperkid123 Looks like the remaining build errors are caused from outdated libs and implementation as I don't recall these being in here before.

```
console.warn
    Warning: The Filterable field has been deprecated on the MultiSelect object. Please import and use FilterableMultiSelect directly.
```

```
DualListSelect › sort value
```

```
DualListSelect › sort options
```

```
console.error
    Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
        at Popover (/home/circleci/react-forms/node_modules/@patternfly/react-core/src/components/Popover/Popover.tsx:216:3)
```

```
console.error
    Warning: React does not recognize the `defaultTime` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `defaulttime` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
        at div
        at div
        at TimePicker (/home/circleci/react-forms/node_modules/@patternfly/react-core/src/components/TimePicker/TimePicker.tsx:88:5)
```

```
console.error
    Warning: Unknown event handler property `onValueChange`. It will be ignored.
        at div
        at Slider (/home/circleci/react-forms/node_modules/@patternfly/react-core/src/components/Slider/Slider.tsx:66:3)
```

```
console.error
    Warning: React does not recognize the `currentValue` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `currentvalue` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
        at div
        at Slider (/home/circleci/react-forms/node_modules/@patternfly/react-core/src/components/Slider/Slider.tsx:66:3)
```

File: packages/react-form-renderer/src/form-renderer/form-renderer.js [hunk 7]
```diff
@@ -1,77 +1,132 @@
...
+  const decoratorsMerged = useMemo(() => ([
+    focusDecorator.current,
+    ...(Array.isArray(decorators) ? decorators : [])
+  ]), [decorators]);
+
+  const handleSubmitCallback = useCallback((values, formApi, ...args) => {
+    return !isFunc(onSubmit) ? undefined : onSubmit(values, {...formApi, fileInputs}, ...args);
+  }, [onSubmit, fileInputs]);
+
+  const handleCancelCallback = useCallback((getState) => (...args) => {
```
Hyperkid123, line 56: @zgover This causes the issue with the unexpected `Cancel` button in FormTemplates.

The common `FormTemplate` is checking if the function exists and if it does, it will render the `cancel` button. As `useCallback` always returns a function regardless of the callback, it confuses the templates.

zgover: @Hyperkid123 I'll leave that to you to propose a solution. It appears there are a few areas that need addressing with the test cancel validation. Personally, I think this should only be validating the `onCancel` pre and post logic, if currently the default behavior displays a cancel element, that should instead be decided by the lib consumer.

Hyperkid123: Yes, they can decide. We are talking about common default templates that are not mandatory to use. So ultimately the decision is on the consumer. I would not like to introduce an entirely new prop that says `enableCancel` or something similar.

You should be able to fix with with a simple `onCancel: onCancel && handleCancelCallback(getState),` assignment

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -53,8 +53,8 @@ const FormRenderer = ({
     return !isFunc(onSubmit) ? undefined : onSubmit(values, {...formApi, fileInputs}, ...args);
   }, [onSubmit, fileInputs]);
 
-  const handleCancelCallback = useCallback((getState) => (...args) => {
-    return !isFunc(onCancel) ? void 0 : onCancel(getState().values, ...args);
+  const handleCancelCallback = useCallback((getState) => {
+    return !isFunc(onCancel) ? undefined : ((...args) => onCancel(getState().values, ...args));
```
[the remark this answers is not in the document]
zgover: @Hyperkid123 Am I missing something? I thought I copied directly almost the exact thing that was there before. The useCallback initial return is called immediately before passing to the context provider.

https://github.com/data-driven-forms/react-forms/blob/3cfbaf6f88ba81a8aae94be2ffd9bcf2c59c1cc0/packages/react-form-renderer/src/form-renderer/form-renderer.js#L85

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -82,11 +137,9 @@ const FormRenderer = ({
               unRegisterInputFile,
               pristine,
               onSubmit,
-              onCancel: onCancel ? (...args) => onCancel(getState().values, ...args) : undefined,
-              onReset: (...args) => {
-                onReset && onReset(...args);
-                reset();
-              },
+              onCancel: handleCancelCallback(getState),
```
Hyperkid123, line 140: The check should be here:
```jsx
onCancel: onCancel && handleCancelCallback(getState),
```

packages/react-form-renderer/src/form-renderer/form-renderer.js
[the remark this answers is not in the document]
Hyperkid123: The issue is that the `onCancel` passed to the renderer is always a function. (use call back will always return a function). Thus the cancel buttons show even if the user did not define any `onCancel` function. Even if you return `undefined` from the callback, it will generate a `() => undefined` function which is still going to trigger the conidtion.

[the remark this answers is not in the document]
Hyperkid123: The implementation of the `handleCallback` was OK before.

zgover:
Seriously, why are these lint checks setup as errors??? No wonder there aren't very many contributions to this repo. It is absolutely far too aggressive, and if that's what's preferred cool, but keep it **preferred** i.e., warn, but not error required, especially for such minor changes. They're not blocking anything besides the build. All of the lint errors caused in this contribution has made my entire experience of reveling my efforts for this lib utterly annoying and does not inspire me to make future updates.

I'd understand it the lint check was on line-endings, multiple empty lines, semicolons, line max column etc... 

But not single line objects, spaces in open/close curly brackets but not square brackets, forced line breaks for lambda/arrow, or forced single line lambda/arrow, disallowing parenthesis to wrap returns/conditional JSX, or shorthands of lambda/arrow functions, array assignments, JSX attribute props, etc.. Merging it all into a single line as required for many of these items limit readability for future maintenance, consumption and overall cleanliness. But if they're left to warnings then reviewers can verify if the warns are ok to approve/merge or if they're outstanding and the code written is ugly/poorly organized then require the lint checks by request.

Hyperkid123:
> They're not blocking anything besides the build. All of the lint errors caused in this contribution has made my entire experience of reveling my efforts for this lib utterly annoying and does not inspire me to make future updates.

You can use the automatic eslint fix command and it will do all the formatting for you. 
```sh
yarn lint --fix
```
(or embedded IDE command)

It took me about 10 seconds to fix all but 3 linting errors and they were related to prop types warnings.

rvsia:
@zgover Do you use an editor or an IDE with ESLINT plugin? To solve all lint errors takes only pressing one keyboard shortcut. Or you can run `yarn lint —fix` command to automatically fix issues. Thanks to prettier, almost all errors can be fixed automatically and we have never heard feedback like this from anyone else yet. And it definitely helps to keep the code cleaner.

zgover:
> @zgover Do you use an editor or an IDE with ESLINT plugin? To solve all lint errors takes only pressing one keyboard shortcut. Or you can run `yarn lint —fix` command to automatically fix issues. Thanks to prettier, almost all errors can be fixed automatically and we have never heard feedback like this from anyone else yet. And it definitely helps to keep the code cleaner.

@rvsia yes I use IntelliJ IDEA for large projects and VSCode for adhoc file viewing. But for very small changes such as this began when starting, I typically make edits in the GitHub web app. Being that it typically takes me more time to pull down a local copy of the branch setup the IDE working space in either one of the editors—all in the amount of time I could make the edits, send an update or PR and be done.

zgover:
> > They're not blocking anything besides the build. All of the lint errors caused in this contribution has made my entire experience of reveling my efforts for this lib utterly annoying and does not inspire me to make future updates.
> 
> 
> 
> You can use the automatic eslint fix command and it will do all the formatting for you. 
> 
> ```sh
> 
> yarn lint --fix
> 
> ```
> 
> (or embedded IDE command)
> 
> 
> 
> It took me about 10 seconds to fix all but 3 linting errors and they were related to prop types warnings.

Lol my point still stands though for such small linting issues they should be warnings, it makes it quick and fluid to make edits in the web app. It's something worth considering being that it's not as though what was written was a mess but rather easier on the eyes for readability.

zgover:
Y'all are more than welcome to partake and not just advise lol, I'm almost positive the option to allow reviewers to edit the PR branch. Particularly, I am not too familiar with the tests in the repo and it would be highly more efficient for y'all to write the last few parts i.e., tests and docs.

Hyperkid123:
> I am not too familiar with the tests in the repo and it would be highly more efficient for y'all to write the last few parts i.e., tests and docs.

We will do the rest of the busy work.

Hyperkid123:
@zgover I had to do one small change to the children as a ReactNode variant to propagate the props. You can check this commit: [fix(renderer): propagate template props to children node](https://github.com/data-driven-forms/react-forms/pull/1218/commits/a6a30b52f579770d51289387aaa2fa7080af69e9)

packages/react-renderer-demo/src/pages/components/children.md
```diff
@@ -0,0 +1,75 @@
...
+*object*
+
+[Schema](/components/renderer#schema) from renderer. You can use it to extract a form title, description, or whatever you need.
+
+## Children node minimal example
+
+Check the [Form template documentation](/components/form-template) for detailed example.
+
+```jsx
+
```
rvsia, line 28: empty line

packages/react-renderer-demo/src/pages/components/renderer.md
```diff
@@ -35,12 +35,24 @@ You can use [globally defined attributes](/mappers/global-component-props).
...
 
 ---
 
+### children
+
+**Note** You have to pass at least one of `FormTemplate` or a `children` render function to render the fields.
+
+*function*
+
+Children render function. It serves the same purpose as `FormTemplate` prop.
```
rvsia, line 52: Can we put a link to children here? 

[Read more](/components/form-template).

rvsia (commented):
Just some small changes to the documentation.

@kp0zhiqian would it be possible to check (even later) changes in the documentation? Thanks! 🙇

packages/react-renderer-demo/src/pages/components/children.md
```diff
@@ -0,0 +1,75 @@
+import DocPage from '@docs/doc-page';
+
+<DocPage>
+
+# Children
+
+[A prop](/components/renderer#children) to customize the form layout. Children render function receive the same props as `FormTemplate`:&nbsp;`formFields` and `schema`.
```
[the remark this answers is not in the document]
kp0zhiqian: emm, I'm not sure what you need me to do? or maybe @ the wrong person?🧐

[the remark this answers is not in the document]
rvsia: Yep, I did, I am sorry 😅 

@yiranc0205 Please could you take a look? 🙏

yiranc0205:
No problem, I will check it later😊

Richard Všianský ***@***.***> 于2022年4月7日周四 16:21写道：

rvsia (approved):

zgover (commented):
@Hyperkid123 was this ever addressed? I navigated away before submitting the changes review with my concerns inside.

packages/react-renderer-demo/src/pages/components/children.md
```diff
@@ -0,0 +1,74 @@
...
+    >
+        <CustomTemplate hideButtons />
+    </FormRenderer>
+  )
+}
+```
+
+## Children render function
+
+Check the [Form template documentation](/components/form-template) for detailed example.
```
yiranc0205, line 58: "example" in the end should be "examples".

packages/react-renderer-demo/src/pages/components/renderer.md
```diff
@@ -35,12 +35,24 @@ You can use [globally defined attributes](/mappers/global-component-props).
 
 *Component*
 
+**Note** You have to pass at least one of `FormTemplate` or a `children` render function to render the fields.
+
 Components which defines a template of the form. This component receives two props from the renderer: `formFields` and `schema`. `formFields` is the content of the form. You should wrap this content into your `<form>` component and add form buttons.
```
yiranc0205, line 40: "defines" should be "define"

PR #1238, opened 2022-04-26
Update children rendering

### LINKED ISSUE #1236: Optimize renderer children props rendering.
This won't work in the following scenarios:

- `children` prop is one of the following: null (typeof null = object), fragment, contexts etc.,
- `children` prop is purely used for adding element to follow the rendered `<FormTemplate/>` and the formFields/schema prop is already reserved within child elements i.e.,
 
```typescript
interface MyInnerElementProps {
  schema: string
}

const MyInnerElement = (props: MyInnerElementProps) => {
  const { schema } = props 
  const { onReset } = useFormApi()

  return (
    <>
      <Head>
        <script 
          type="application/ld+json"
          dangerouslySetInnerHTML={{ __html: schema }}
        />
      </Head>
      <button onClick={() => onReset()}/>
        Reset
      </button>
    </>
  )
}

<FormRenderer
  FormTemplate={MyFormTemplate}
  // ...
>
  <MyInnerElement
    schema={`
{
  "@context": "https://json-ld.org/contexts/person.jsonld",
  "@id": "http://dbpedia.org/resource/John_Lennon",
  "name": "John Lennon",
  "born": "1940-10-09",
  "spouse": "http://dbpedia.org/resource/Cynthia_Lennon"
}
`}
  />
</FormRenderer>
```

We could do something like this if we have to clone and forward props to element, but second problem still exists unless they've passed a Fragment is the first child or similar:

```typescript

const renderChildren = () => {
  if (isFunc(children)) {
    return children({ formFields, schema })
  }
  if (ReactIs.isElement(children)) {
    return cloneElement(children, { formFields, schema })
  }
  
  return children
}

renderChildren()

```

_Originally posted by @zgover in https://github.com/data-driven-forms/react-forms/pull/1218#discussion_r843847625_

Hyperkid123:
Fixes #1236

**Description**
Add additional options to render children in form renderer

- allow passing `null` as a child
- only permit one root child element
- use `schema` and `formFields` if explicitly given to renderer child
- expose schema in `formOptions`

- [x] `Yarn build` passes
- [x] `Yarn lint` passes
- [x] `Yarn test` passes
- [x] Correct commit message
   - scope: package

### DIFF
--- a/packages/react-form-renderer/src/form-renderer/form-renderer.js
+++ b/packages/react-form-renderer/src/form-renderer/form-renderer.js
@@ -12,6 +12,33 @@ import SchemaErrorComponent from './schema-error-component';
 
 const isFunc = (fn) => typeof fn === 'function';
 
+const renderChildren = (children, props) => {
+  if (isFunc(children)) {
+    return children(props);
+  }
+
+  let childElement = children;
+  if (Array.isArray(children)) {
+    /**
+     * Only permit one child element
+     */
+    if (children.length !== 1) {
+      throw new Error('FormRenderer expect only one child element!');
+    }
+
+    childElement = children[0];
+  }
+
+  if (typeof childElement === 'object') {
+    /**
+     * Clone react element, pass form fields and schema as props, but override them with child props if present
+     */
+    return cloneElement(children, { ...props, ...childElement.props });
+  }
+
+  throw new Error(`Invalid children prop! Expected one of [null, Function, object], got ${typeof children}`);
+};
+
 const FormRenderer = ({
   actionMapper,
   children,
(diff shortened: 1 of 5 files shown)


Hyperkid123:
@zgover this should resolve the issue you have pointed out from #1236

packages/react-form-renderer/src/form-renderer/form-renderer.js
```diff
@@ -144,13 +175,13 @@ const FormRenderer = ({
...
               initialValues: props.initialValues,
+              schema,
             },
           }}
         >
           {FormTemplate && <FormTemplate formFields={formFields} schema={schema} {...FormTemplateProps} />}
 
-          {isFunc(children) && children({ formFields, schema })}
-          {typeof children === 'object' && cloneElement(children, { formFields, schema })}
+          {renderChildren(children, { formFields, schema })}
```
rvsia, line 184: could we get something like:

`children && render(...` 

It will remove the calling of the function and from the function the null/undefined can be removed.

packages/react-renderer-demo/src/pages/hooks/use-form-api.md
```diff
@@ -26,6 +26,12 @@ This hook returns object containing following information:
 
 If you want to render fields from a component (`tabs`, `subform`, etc.) you can use `renderForm(fields)` function.
 
+## schema
+
+*object*
+
+The schema prop of FormRenderer.
```
rvsia, line 33: Can we get a link to schema here? 

https://data-driven-forms.org/components/renderer#schema

PR #1251, opened 2022-05-12
Improve rendering performance

### LINKED ISSUE #1243: [V4] Validation onChange is too slow
**Scope**: Manager

**Description**

In demo schema onChange validation causes issues (some memoization? see TODO)

Hyperkid123:
Fixes #1243

**Description**

- Introduce validation cache to improve validation performance
- remove deep clone from render transition
- manually set render events

### DIFF
--- a/packages/form-state-manager/src/manager-api/manager-api.ts
+++ b/packages/form-state-manager/src/manager-api/manager-api.ts
@@ -873,7 +873,7 @@ const createManagerApi: CreateManagerApi = ({
         validateForm(config.validate);
       }
 
-      render(['values', 'errors']);
+      render(['values', 'errors', 'valid', 'invalid']);
     });
   }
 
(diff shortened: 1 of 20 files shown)


packages/form-state-manager/src/manager-api/manager-api.ts
```diff
@@ -516,45 +518,92 @@ const createManagerApi: CreateManagerApi = ({
...
+            valid: true,
+            invalid: false,
+            validating
+          }}
+          validationCache.set(cacheKey, newState)
+        return newState});
     }
   }
 
+  function getCacheKey(keyTemplate: any): string {
```
rvsia, line 565: This function does not use anything of the internal state - we can turn it to an external function

packages/form-state-manager/src/tests/performance/field-render-cycle.test.js
```diff
@@ -2,7 +2,7 @@ import React, { useEffect, useContext, Fragment } from 'react';
 import { act, render as rtlRender, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
-import useField from '../../use-field';
+import useField from '../../use-field/use-field';
```
rvsia, line 5: 👁️

rvsia (approved):

PR #1261, opened 2022-05-31
Use node circle orb

Hyperkid123:
Removes outdated docker images from circle CI config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,16 +1,17 @@
 version: 2.1
 orbs:
   codecov: codecov/codecov@1.0.4
+  node: circleci/node@5.0.2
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,16 +1,19 @@
...
     steps:
       - checkout
-      - run:
-          name: update-npm
-          command: 'sudo npm install -g npm@latest'
+      - node/install:
+          install-yarn: true
+          node-version: '16.13'
+      - run: node --version
+      - run: npm --version
```
rvsia, line 16: Do we need to run these?

rvsia: same below

PR #1267, opened 2022-06-06
[MANAGER] Fix(renderer): use types from manager

### LINKED ISSUE #935: [v3] Remove final form dependency and fix types
**Scope**: renderer

**Description**

Remove all mentions of react-final-form from renderer

rvsia:
Fixes #935

**Description**

Changes to the usage of manager types, not DDF's.

### DIFF
--- a/packages/form-state-manager/src/use-field/use-field.ts
+++ b/packages/form-state-manager/src/use-field/use-field.ts
@@ -49,17 +49,17 @@ export interface Meta {
   warning: any;
 }
 
-export type OnChange = (value: any) => void;
-export type OnBlur = (value: any) => void;
-export type OnFocus = (value: any) => void;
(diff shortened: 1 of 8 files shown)


packages/form-state-manager/src/form-state-manager/form-state-manager.tsx
```diff
@@ -12,7 +12,7 @@ import { Debug } from '../manager-api';
 
 export interface FormStateManagerProps {
   onSubmit: (values: AnyObject) => void;
-  children?: (props: AnyObject) => React.ReactNode;
+  children?: (props: AnyObject) => React.ReactNode | React.ReactNode;
```
Hyperkid123, line 15: DO we need the same type twice?

Hyperkid123 (approved):

PR #1299, opened 2022-08-04
fix(carbon): do not use absolute import paths

### LINKED ISSUE #1298: Upgrading to carbon-components-react 7.56 causing components to break
After upgrading to carbon-components-react 7.56 I ran into this bug with the Carbon DDF dual list select component.
```

Uncaught (in promise) Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
Check the render method of DualListSelectInner
```

Not sure if this affects other components but this is the only one I found.

Hyperkid123:
There are import path changes in minor releases. Using absolute imports
will cause backwads compatibility issues.

Fixes #1298

### DIFF
--- a/packages/mui-component-mapper/src/date-picker/date-picker.d.ts
+++ b/packages/mui-component-mapper/src/date-picker/date-picker.d.ts
@@ -6,6 +6,10 @@ type InternalDatePickerProps<TInputDate, TDate> = MuiDatePickerProps<TInputDate,
 }
 
 export type DatePickerProps<TInputDate, TDate> = InternalDatePickerProps<TInputDate, TDate> & UseFieldApiComponentConfig;
+<<<<<<< Updated upstream
+=======
+
+>>>>>>> Stashed changes
 declare const DatePicker: React.ComponentType<DatePickerProps<any, any>>;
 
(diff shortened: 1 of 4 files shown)


rvsia (approved):

packages/mui-component-mapper/src/date-picker/date-picker.d.ts
```diff
@@ -6,6 +6,10 @@ type InternalDatePickerProps<TInputDate, TDate> = MuiDatePickerProps<TInputDate,
 }
 
 export type DatePickerProps<TInputDate, TDate> = InternalDatePickerProps<TInputDate, TDate> & UseFieldApiComponentConfig;
+<<<<<<< Updated upstream
```
GilbertCherrie, line 9: @Hyperkid123 Should these be removed? I think these are the git comments but might cause parsing errors.

Hyperkid123: Oh yeah. Nice catch!

PR #1322, opened 2022-09-10
Checkbox validation fix

jspannu919:
Issue:

For checkbox component, if a "required" validation is passed then in that case a validation error is thrown if the user deselects the value. This shouldn't be the case because "false" is a valid value for boolean data.

### DIFF
--- a/packages/react-form-renderer/src/tests/validators/validators.test.js
+++ b/packages/react-form-renderer/src/tests/validators/validators.test.js
@@ -10,6 +10,14 @@ describe('New validators', () => {
       expect(validatorMapper[validatorTypes.REQUIRED]()('Foo')).toBeUndefined();
     });
 
+    it('should pass required validator with false value', () => {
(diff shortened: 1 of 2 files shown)


rvsia:
@jspannu919 Thanks for this PR, however, this is not going to work.

`required` in context od switches/checkboxes is being used for confirmations. The most common case is "I agree with product terms".

`False` is and should be considered as an empty state.

Hyperkid123:
@jspannu919 the `null` value is sometimes used in checkboxes that have `tristate`. However that does not follow HTML spec and is usually built into the component itself.

By default, the checkbox value is defined as a `boolean`. We have to follow the HTML standards.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#attr-checked

For non-standard behavior you sould create a custom component that will send the `null` directly in the `onChange` event.

jspannu919:
Okay, thanks guys! Will override the validation mapper locally for fixing my issue.

PR #1323, opened 2022-09-10
Float Issue fix

jspannu919:
When typing a "." in a field having dataType as "float", the input is ignored and due to which user is never able to type a decimal value even if the type is set to "float". 

Reason: The canBeConvertedToNumber functions returns true for values like "11.", but the parseFloat functions returns "11" only and thus removing the decimal.

Issue is replicated at https://stackblitz.com/edit/1dkk8s?file=schema.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -189,7 +189,5 @@
       ]
     ]
   },
-  "dependencies": {
-    "yarn": "^1.22.19"
-  }
+  "dependencies": {}
 }
--- a/package.json
+++ b/package.json
@@ -189,5 +189,7 @@
       ]
     ]
   },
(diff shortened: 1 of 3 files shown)


Hyperkid123 (approved):
@jspannu919 Looks great. We need a correct commit message format to trigger automatic release. We use the classic semver `feat|fix(package): Message`.

I am ok to chaning the commits, but don't want to push to your branch before asking

Hyperkid123:
@jspannu919 the CI is a bit flaky ATM. I am working on a fix. (has to do with react 17/18). I have restarted the failed job.

PR #1331, opened 2022-10-04
feat(renderer): enable condition set as a function

### LINKED ISSUE #1326: Allow `set` condition attribute to be a function
Hi, before my question I want to thank you for your great library and your efforts to support and keep it updated.

I have a custom field array which its default values are set from other components based on some conditions. For example, in this sandbox, it is based on the text field if the value given is '3'.
It works fine if I want to set/replace the whole field array with a given value but how can I append to it?

in the example provided, the field array initial values are 1 -> one and 2 -> two or whatever user type, and I want to append the 3->three if I write 3 in the text field, how can I achieve this?

Code Exaple: [codesandbox](https://codesandbox.io/s/append-to-fieldarray-i5edcs?file=/src/App.jsx)

Thanks

_Originally posted by @Ali-Zmn in https://github.com/data-driven-forms/react-forms/discussions/1319_

### Required format:

The `set` attribute should be either an object or a function that returns desired object.

```TS
type set = object | (formApi: FormApi) => object
```

Function result should be extracted here and added to the state: https://github.com/data-driven-forms/react-forms/blob/master/packages/react-form-renderer/src/condition/condition.js#L48

JuliePrazakova:
Fixes #1326 

- [x] `Yarn build` passes
- [x] `Yarn lint` passes
- [x] `Yarn test` passes
- [x] Correct commit message
   - scope: package

### DIFF
--- a/packages/react-form-renderer/src/condition/condition.js
+++ b/packages/react-form-renderer/src/condition/condition.js
@@ -1,9 +1,19 @@
-import React, { useEffect, useReducer } from 'react';
+import React, { useCallback, useEffect, useReducer } from 'react';
 import PropTypes from 'prop-types';
 import isEqual from 'lodash/isEqual';
 
 import useFormApi from '../use-form-api';
 import parseCondition from '../parse-condition';
+
+const setterValueCheck = (setterValue) => {
+  if (setterValue === null || Array.isArray(setterValue)) {
+    console.error('Received invalid setterValue. Expected object, received: ', setterValue);
+
+    return false;
+  }
+
+  return typeof setterValue === 'object';
+};
 
 export const reducer = (state, { type, sets }) => {
   switch (type) {
@@ -42,6 +52,12 @@ const Condition = React.memo(
       }
     }, [dirty]);
 
+    const setValue = useCallback((setter) => {
+      Object.entries(setter).forEach(([name, value]) => {
+        formOptions.change(name, value);
+      });
+    }, []);
+
     useEffect(() => {
       if (setters && setters.length > 0 && (state.initial || !isEqual(setters, state.sets))) {
         setters.forEach((setter, index) => {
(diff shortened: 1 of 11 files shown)


packages/react-form-renderer/src/condition/condition.js
```diff
@@ -42,6 +52,14 @@ const Condition = React.memo(
       }
     }, [dirty]);
 
+    const x = useCallback(
```
rvsia, line 55: `x` is kinda a mysterious name 😵‍💫

packages/react-form-renderer/src/condition/condition.js
```diff
@@ -42,6 +52,14 @@ const Condition = React.memo(
...
+      (setter) => {
+        Object.entries(setter).forEach(([name, value]) => {
+          formOptions.change(name, value);
+        });
+      },
+      [formOptions.change]
+    );
     useEffect(() => {
```
rvsia, line 63: Can you please put an empty line above ^

packages/react-form-renderer/src/tests/form-renderer/condition.test.js
```diff
@@ -531,6 +531,150 @@ describe('condition test', () => {
     await waitFor(() => expect(screen.getByLabelText('field2')).toHaveValue('set with then'));
   });
 
+  test('should change fileds value by set funciton', async () => {
```
rvsia, line 534: please use `it` instead of `test`, also a typo her

packages/react-form-renderer/src/condition/condition.js
```diff
@@ -42,6 +52,12 @@ const Condition = React.memo(
       }
     }, [dirty]);
 
+    const setValue = useCallback((setter) => {
+      Object.entries(setter).forEach(([name, value]) => {
+        formOptions.change(name, value);
+      });
+    });
```
rvsia, line 59: We still need to keep an empty dependency array here: `[]` otherwise the function will be always re-created.

See [here](https://beta.reactjs.org/apis/react/useCallback#every-time-my-component-renders-usecallback-returns-a-different-function)

rvsia (approved):

PR #1344, opened 2022-12-02
Bump editor version in docs.

Hyperkid123:


rvsia:
@Hyperkid123 in version 14 default usage of MUI was removed.

We need to manually use https://github.com/data-driven-forms/editor/blob/main/packages/editor-pro/src/editor/sub-form.tsx component instead of the default one from the mapper. It should be possible to just import it.

In the future, we have to do it more flexible. (Some customizable sub-form common component.)

PR #1369, opened 2023-02-20
Remove obsolete firebase deployment.

Hyperkid123:
Removes firebase dependencies.


packages/react-renderer-demo/firebaseFunctions.js
```diff
@@ -1,96 +0,0 @@
...
-      nextjsServer.serveStatic(req, res, filePath);
-    } else {
-      nextjsHandle(req, res, parsedUrl);
-    }
-  });
-});
-
-const github = require('@actions/github');
-
-exports.sendComment = functions.https.onRequest(async (request, response) => {
```
rvsia, line 66: @Hyperkid123 We should keep this part (and migrate it to somewhere else) as it handles the bot.

Is it possible to migrate it to vercel as well? 🤔

rvsia: https://vercel.com/docs/concepts/functions/serverless-functions

Hyperkid123: I can't do it now I am on meetings whole week. If you create an issue I can work on it later. I just want the build to stop failing.

rvsia: Yeah, actually we can remove it - as long we won't remove anything from the firebase itself, it should be fine.

rvsia (approved):

Hyperkid123:
@rvsia its running now. There was still a reference to the deploy job in circle ci config.

PR #1380, opened 2023-04-28
fix(mui): Apply SwitchProps to Switch component

danielwarke:
**Description**

In the current code, the SwitchProps are not being applied to the Switch component. We should apply these SwitchProps so that users can customize the behavior of the component as this was the original intent.

- [x] `Yarn build` passes
- [x] `Yarn lint` passes
- [x] `Yarn test` passes


rvsia:
@danielwarke thank you for this pr! 😍 

This PR should be a fix, as the original behavior is a bug. Could you please change the commit message (and PR title) to `fix(mui): …`?

danielwarke:
> @danielwarke thank you for this pr! 😍
> 
> This PR should be a fix, as the original behavior is a bug. Could you please change the commit message (and PR title) to `fix(mui): …`?

No problem! 👍🏻 I've updated the PR title and description to "fix"

rvsia:
@danielwarke we need to do it also for the commit message, otherwise no new version would be released: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message

PR #1381, opened 2023-04-28
fix(mui): apply SwitchProps to Switch component

danielwarke:
**Description**

In the current code, the SwitchProps are not being applied to the Switch component. We should apply these SwitchProps so that users can customize the behavior of the component as this was the original intent.

- [x] `Yarn build` passes
- [x] `Yarn lint` passes
- [x] `Yarn test` passes


danielwarke:
@rvsia Here is the recreated PR with `fix` prefix 🙏🏻

rvsia (approved):

PR #1382, opened 2023-05-20
Fix numeric validator

jspannu919:
Currently the numeric validators (min or max) doesn't shows any validation error if the value in the field is zero.

Scenario:
If a field is added with min validation with threshold as 3, the validator will show error for values like 1,2,-1 etc but nothing for 0.

### DIFF
--- a/packages/react-form-renderer/src/validators/validator-functions.js
+++ b/packages/react-form-renderer/src/validators/validator-functions.js
@@ -90,7 +90,7 @@ export const numericality = memoize(
     lessOrEqual = selectNum(lessOrEqual, lessThanOrEqualTo);
 
     return prepare((value) => {
-      if (!value) {
+      if (value == null) {
         return;
       }
 
--- a/packages/react-form-renderer/src/validators/validator-functions.js
+++ b/packages/react-form-renderer/src/validators/validator-functions.js
@@ -90,7 +90,7 @@ export const numericality = memoize(
     lessOrEqual = selectNum(lessOrEqual, lessThanOrEqualTo);
 
     return prepare((value) => {
-      if (value == null) {
+      if (value === null || value === undefined) {
         return;
       }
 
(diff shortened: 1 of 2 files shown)


packages/react-form-renderer/src/validators/validator-functions.js
```diff
@@ -90,7 +90,7 @@ export const numericality = memoize(
     lessOrEqual = selectNum(lessOrEqual, lessThanOrEqualTo);
 
     return prepare((value) => {
-      if (!value) {
+      if (value == null) {
```
Hyperkid123, line 93: I think we should cover all "falsy" values
``` suggestion
      if (typeof value !== 'number' && typeof value !== 'string') {
```

Hyperkid123 (commented):
@jspannu919 Thanks for the contribution. I have one small suggestion.

packages/react-form-renderer/src/validators/validator-functions.js
jspannu919: The next block to this is checking if the value is a valid number or not, handling these here will lead to no proper message being displayed.

Hyperkid123: You are correct. My point is that we also have to consider `undefined` in addition to `null`. I guess we can let the rest of the types through this check.

jspannu919: I have not added strict equality intentionally, to cover "null" as well as "undefined" case.

Hyperkid123: These strings will be handled by the `isNumber` check below as you pointed out. We like to use the strict equality operator whenever possible to avoid unexpected "truthy" results. It's a matter of convention.

We can go ahead and merge/release the PR afterward.

jspannu919: Sorry if I was not able to explain it properly. I didn't meant to cover the null and undefined "strings" but the values, so basically to cover (value === null || value === undefined) with just ( value == null ) as it will return true for undefined also. Let me know if you think we should explicitly check for undefined with strict equality, I will make the change.

Hyperkid123: @jspannu919 yes, please. We try to use strict equality as much as possible to avoid some unintentional condition outcomes.

jspannu919: Done

Hyperkid123 (approved):

PR #1384, opened 2023-06-02
Fix data type validation message for numeric validator in case of custom message

jspannu919:
Currently if have a validation (ex: minimum) on a numeric field and the value is not a number then it shows the data type wrong message (Value must be an integer) and if the type is correct then it shows the correct message (Value must be greater or equal.....). But when we are passing a custom validation message, then the data type message is not displayed and in that case also the custom message is shown.

### DIFF
--- a/packages/react-form-renderer/src/tests/validators/validators.test.js
+++ b/packages/react-form-renderer/src/tests/validators/validators.test.js
@@ -133,9 +133,9 @@ describe('New validators', () => {
     });
 
     it('should not pass the validation and return type error message', () => {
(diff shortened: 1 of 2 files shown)


rvsia (approved):
Thank you for this PR! 

Looks good to me, could you please solve the conflict? 😊

rvsia:
@Hyperkid123 the bot is showing that no version will be released.. do you see anything wrong here or is the bot broken? 🤔

jspannu919:
> Thank you for this PR!
> 
> Looks good to me, could you please solve the conflict? 😊

Done. Got conflicts with my own changes ;-)

PR #1387, opened 2023-06-05
fix(renderer): nested condition based on form initial values

### LINKED ISSUE #1383: Conditional statement not working for form using initialValues prop of FormRenderer
I found one issue where the form has initialValues set using prop of FormRenderer and I'm using clearOnMount and initializeOnMount as true for all the fields. Now when I go to some other subForm and come back to the first subForm, all the fields get their initialValues set but it fails to show fields based on radio button condtions even though radio button is set to the  expected initial value. 

Sharing an example replicating the issue:
[https://stackblitz.com/edit/ccpsrt?file=schema.js,index.js](https://stackblitz.com/edit/ccpsrt?file=schema.js,index.js)

In example, you can see the data in last text-field for the first time. But when you select another radio button (form1 suppose) and then come back to form2, last text-field won't come which should be there based on radio button selected condition.

Hyperkid123:
Fixes #1383

**Description**

The input of type `radio` does not trigger the condition after `re-mount` if the condition is based on `initialValues` renderer prop.

Issue is with the `values` that are collected from all the triggers. Because the `radio` type is registered multiple times, per each option, the `useField` does not `yield` the actual state value immediately. Therefore we have to use the `getState().values` to retrieve the state values and pass it to the condition parser instead of the smaller, `values` collected from the triggers. 

To reproduce the issue, use the following schema with initial values. Select `form1` from the top radio component, and then `reselect` the `form2` option.

Without the fix the `Radio1` text field will not be visible, even though the values for the condition are correct.

**Schema** *(if applicable)*

```jsx
  const initialValues = {
    "formRadio": "form2",
    "radioBtn2": "stu",
    "txtField3": "data"
  }
  const scema =  {
    "fields": [
      {
        "name": "formRadio",
        "label": "SelectSubForm",
        "component": "radio",
        "initializeOnMount": true,
        "clearOnUnmount": true,
        "options": [
          { "label": "form1", "value": "form1" },
          { "label": "form2", "value": "form2" }
        ]
      },
      {
        "component": "sub-form",
        "name": "subform1",
        "condition": {
          "when": "formRadio",
          "is": "form1"
        },
        "fields": [
          {
            "name": "txt1",
            "label": "Enter text",
            "initializeOnMount": true,
            "clearOnUnmount": true,
            "component": "text-field"
          },
          {
            "name": "radioBtn",
            "label": "Select",
            "component": "radio",
            "initializeOnMount": true,
            "clearOnUnmount": true,
            "options": [
              { "label": "abc", "value": "abc" },
              { "label": "def", "value": "def" }
            ]
          },
          {
            "name": "txtField2",
            "label": "Radio1",
            "initializeOnMount": true,
            "clearOnUnmount": true,
            "component": "text-field",
            "condition": {
              "when": "radioBtn",
              "is": "def"
            }
          }
        ]
      },
      {
        "component": "sub-form",
        "name": "subform2",
        "condition": {
          "when": "formRadio",
          "is": "form2"
        },
        "fields": [
          {
            "name": "radioBtn2",
            "label": "Select",
            "component": "radio",
            "initializeOnMount": true,
            "clearOnUnmount": true,
            "options": [
              { "label": "pqr", "value": "pqr" },
              { "label": "stu", "value": "stu" }
            ]
          },
          {
            "name": "txtField3",
            "label": "Radio1",
            "initializeOnMount": true,
            "clearOnUnmount": true,
            "component": "text-field",
            "condition": {
              "and": [
                {
                  "when": "radioBtn2",
                  "is": "stu"
                },
                {
                  "when": "formRadio",
                  "is": "form2"
                }
              ]
            }
          }
        ]
      }
    ]
  }
```

### DIFF
--- a/packages/react-form-renderer/src/condition/condition.js
+++ b/packages/react-form-renderer/src/condition/condition.js
@@ -43,7 +43,10 @@ const Condition = React.memo(
       initial: true,
     });
 
-    const conditionResult = parseCondition(condition, values, field);
+    // It is required to get the context state values from in order to get the latest state.
+    // Using the trigger values can cause issues with the radio field as each input is registered separately to state and does not yield the actual field value.
+    const conditionResult = parseCondition(condition, formOptions.getState().values, field);
+
     const setters = conditionResult.set ? [conditionResult.set] : conditionResult.sets;
 
     useEffect(() => {
--- a/packages/react-form-renderer/demo/form-fields-mapper.js
+++ b/packages/react-form-renderer/demo/form-fields-mapper.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-unused-vars */
 import React from 'react';
-import { componentTypes } from '../src';
+import { componentTypes, useFormApi } from '../src';
 import FieldProvider from '../src/field-provider';
 import useFieldApi from '../src/use-field-api';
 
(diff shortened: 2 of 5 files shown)


rvsia (approved):

packages/react-form-renderer/src/condition/condition.js
```diff
@@ -43,7 +43,10 @@ const Condition = React.memo(
       initial: true,
     });
 
-    const conditionResult = parseCondition(condition, values, field);
+    // It is required to get the context state values from in order to get the latest state.
+    // Using the trigger values cna cause issues with the radio field as each input is registered separately to state and does not yield the actual field value.
```
rvsia, line 47: typo in comment 'cna' > 'can'

PR #1391, opened 2023-06-23
fix(condition): Remove useMemo and parse condition on change

### LINKED ISSUE #1390: Additional field props are not update for conditional fields

Form renderer

**Description**

Here at [NorthStar](https://github.com/aws/aws-northstar), we have [custom implementation of various component mappers](https://github.com/aws/aws-northstar/tree/main/packages/ui/src/components/FormRenderer). 

For the [Wizard component mapper](https://github.com/aws/aws-northstar/blob/main/packages/ui/src/components/FormRenderer/components/Wizard/index.tsx), the expected behavior is the validation should be triggered when users click the Next button for each step. We have incorporated an additional flag/props called showError in each component mapper to facilitate this requirement. When showError is set to true and there is error, the error will be displayed. The showError flag will be set to true when users click Next and will be propagated into each components. 

This approach works for fields without conditions. However, we noticed it does not work for fields with condition set.  

You can find a CodeSandbox example [here](https://codesandbox.io/s/sharp-bogdan-578mzc). Steps to reproduce the behavior:
1. Toggle on Show conditional input
2. Press Next on the wizard without filling any fields
3. The wizard does not go to the next step - validation is performed but validation errors for conditional components are not shown
    *The console log shows the form invalid=true. However, the condition Field was not updated at all while another fields are updated with the showError values. The `fields` props did include all fields*
4. Fill in required fields and click Next. Next page in the wizard is displayed

I have debugged from my end, however, without luck. Raise this issue to see whether you have better approach to implement this feature for the condition fields?

rvsia:
Fixes #1390 

**Description**

- React.memo was caching the field based on condition/values. I rewrote the logic to not cache the whole component but only the result of the parse condition function

Kapture 2023-06-23 at 13 28 37

Performance should stay unaffected as well

**Schema** *(if applicable)*

see issue

### DIFF
--- a/packages/react-form-renderer/src/condition/condition.js
+++ b/packages/react-form-renderer/src/condition/condition.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useReducer } from 'react';
+import { useCallback, useEffect, useMemo, useReducer } from 'react';
 import PropTypes from 'prop-types';
 import isEqual from 'lodash/isEqual';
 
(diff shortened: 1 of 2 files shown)


rvsia:
@jessieweiyi would you please able to test this fix locally and give here an approval if everything is fine? as @Hyperkid123 is now on vacation, I would like to get some approval before merging, thank you! 😊

jessieweiyi:
> @jessieweiyi would you please able to test this fix locally and give here an approval if everything is fine? as @Hyperkid123 is now on vacation, I would like to get some approval before merging, thank you! 😊

Hi @rvsia. Sure. Will do when I am available. Probably will be tmr morning Australian daytime.

packages/react-form-renderer/src/condition/condition.js
```diff
@@ -33,76 +33,73 @@ export const reducer = (state, { type, sets }) => {
...
+    sets: [],
+    initial: true,
+  });
 
-    // It is required to get the context state values from in order to get the latest state.
-    // Using the trigger values can cause issues with the radio field as each input is registered separately to state and does not yield the actual field value.
-    const conditionResult = parseCondition(condition, formOptions.getState().values, field);
+  // It is required to get the context state values from in order to get the latest state.
+  // Using the trigger values can cause issues with the radio field as each input is registered separately to state and does not yield the actual field value.
+  const conditionResult = useMemo(() => parseCondition(condition, formOptions.getState().values, field), [formOptions.getState().values, condition]);
```
jessieweiyi, line 47: `field` is used inside the code block and should it be listed as one of the dependencies?

rvsia: Yeah, I was thinking about that a lot. We use `field` there to only pass `field.name` to conditions in case of dynamic forms and `name` cannot be changed anyways.

However, I added there and then tested it - performance still looks fine, so fixed. Thank you for pointing this out! 😊

jessieweiyi (approved):

jessieweiyi:
Thanks for the fix. I have tested the changes in my local and it works fine for our use cases.

Hyperkid123:
@rvsia we might have to run some benchmarks with this change. I am wondering what impact this will have on forms with a large number of conditional fields.

PR #1410, opened 2023-09-06
Add compareValues prop to useSelect's reducer

### LINKED ISSUE #1401: Options with "value" field as object or array type are duplicated
**Scope**: PF4 mapper, Form renderer

**Description**

How to reproduce: use a schema with `component` set to "select", and `isSearchable` to "true" with `loadOptions` function (so that you are able to asynchronously load values for the select component. Also, in `loadOptions` return a promise that resolves in an array with options. The options have to have value of array or object type, not primitive (like string or numbers).

Current behavior: the select component duplicates the value as you type in the select component and new data loads.

[Screencast from 2023-08-22 18-01-52.webm](https://github.com/data-driven-forms/react-forms/assets/31385370/b656cb15-7545-4aff-8903-a133e4abb60d)

Expected behavior: the select component renders just only the options it gets from `loadOptions`.

**Schema**

```
const loadOptions = (searchValue = '') => {
  return getGroups({ name: searchValue }).then((data) => {
    return data.results.map(({ name, id }) => ({
      label: name,
      value: { id, name },
    }));
  });
};

const exampleSchema = (hosts) => ({
  fields: [
    {
      component: 'select',
      name: 'group',
      label: 'Select a group',
      simpleValue: true,
      isSearchable: true, // enables typeahead
      isRequired: true,
      isClearable: true,
      placeholder: 'Type or click to select a group',
      loadOptions,
      options: [],
      validate: [{ type: validatorTypes.REQUIRED }],
    }
  ],
});
```

victoria-dos:
Fixes #1401

**Description**

**Schema** *(if applicable)*

```jsx
const loadOptions = (searchValue = '') => {
  return getGroups({ name: searchValue }).then((data) => {
    return data.results.map(({ name, id }) => ({
      label: name,
      value: { id, name },
    }));
  });
};

const exampleSchema = (hosts) => ({
  fields: [
    {
      component: 'select',
      name: 'group',
      label: 'Select a group',
      simpleValue: true,
      isSearchable: true, // enables typeahead
      isRequired: true,
      isClearable: true,
      placeholder: 'Type or click to select a group',
      loadOptions,
      options: [],
      validate: [{ type: validatorTypes.REQUIRED }],
    }
  ],
});
```

- [x] `Yarn build` passes
- [x] `Yarn lint` passes
- [x] `Yarn test` passes
- [x] Correct commit message
   - scope: package

### DIFF
--- a/packages/common/src/select/select.js
+++ b/packages/common/src/select/select.js
@@ -4,6 +4,7 @@ import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import clsx from 'clsx';
 import useSelect from '../use-select/use-select';
+import deepEqual from './deep-equal';
 
 const Select = ({
   invalid,
@@ -24,6 +25,7 @@ const Select = ({
   SelectComponent,
   noValueUpdates,
   optionsTransformer,
+  compareValues = deepEqual,
   ...props
 }) => {
   const {
@@ -44,6 +46,7 @@ const Select = ({
     pluckSingleValue,
     isMulti,
     simpleValue,
+    compareValues,
   });
 
   const renderNoOptionsMessage = () => (Object.values(state.promises).some((value) => value) ? () => updatingMessage : () => noOptionsMessage);
@@ -119,6 +122,7 @@ Select.propTypes = {
   SelectComponent: PropTypes.elementType.isRequired,
   noValueUpdates: PropTypes.bool,
   optionsTransformer: PropTypes.func,
+  compareValues: PropTypes.func,
 };
 
 Select.defaultProps = {
@@ -129,6 +133,7 @@ Select.defaultProps = {
   placeholder: 'Choose...',
   isSearchable: false,
   isClearable: false,
+  compareValues: deepEqual,
 };
 
 export default Select;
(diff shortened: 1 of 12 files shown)


packages/common/src/tests/select/select.test.js
```diff
@@ -501,7 +502,149 @@ describe('Select test', () => {
...
+                },
+              ],
+            }}
+          />
+        );
+      });
+
+      // screen.debug(screen.getByRole('form'));
+
+      // expect(state.value).toEqual([{ label: 'first' }]);
```
rvsia, line 537: comment leftovers

packages/common/src/select/select.js
```diff
@@ -5,6 +5,36 @@ import PropTypes from 'prop-types';
 import clsx from 'clsx';
 import useSelect from '../use-select/use-select';
 
+const deepEqual = (a, b, depth = 0) => {
```
rvsia, line 8: Could we provide some unit tests for this function and extract it to a separate file?

Hyperkid123: Sure

rvsia (approved):

PR #1424, opened 2023-11-07
feat(renderer): allow condition mapping

### LINKED ISSUE #1418: Question: Is it possible to use a custom function for conditions within the schema in JSON format?
**Scope**: react-form-renderer / conditions

**Description**

I am looking for a way to define a condition via custom function (https://www.data-driven-forms.org/schema/is#customfunction) BUT i cannot use a javascript object, the schema has to be in JSON format since it will be retrieved from the backend. 

I've looked into the ActionMapper (https://www.data-driven-forms.org/mappers/action-mapper), but as far as I understand, the mapped functions can only be applied to a field property. In my case I would need to assign a mapped function inside a (possibly nested) condition - some sort of "ConditionMapper" if that makes any sense. See example below.

My question is: Can this somehow be done with the current implementation, or would this be a feature request?

**Schema**

This is the javascript version, works as documented:

```
const schema = {
  title: 'Is condition',
  fields: [
    {
      name: 'field-1',
      label: 'Field 1'
    },
    {
      name: 'field-2',
      label: 'Field 2',
      condition: { when: 'field-1', is: (value, config) => someCustomLogic(value, config) },
    },
  ],
};
```

This would be the corresponding JSON-Version, I have no idea if/how this can be done at the moment:

```
const schemaJSON = {
  "title": "Is condition",
  "fields": [
    {
      "component": "text-field",
      "name": "field-1",
      "label": "Field 1"
    },
    {
      "component": "text-field",
      "name": "field-2",
      "label": "Field 2"
      "condition": { "when": "field-1", is: ["??? somehow mapped function ???"] },
    },
  ],
};
```

Hyperkid123:
fixes #1418

### DIFF
--- a/packages/react-form-renderer/demo/index.js
+++ b/packages/react-form-renderer/demo/index.js
@@ -9,6 +9,22 @@ import mapper from './form-fields-mapper';
 const schema = {
   fields: [
     {
+      name: 'field1',
+      label: 'Field 1',
+      component: 'text-field',
+    },
+    {
+      name: 'mapped-condition',
+      label: 'Mapped Condition',
+      component: 'text-field',
+      condition: {
+        mappedAttributes: {
+          is: ['nameFn', 'John', 'Doe'],
+        },
+        when: 'field1',
+      },
+    },
+    {
       name: 'formRadio',
       label: 'SelectSubForm',
       component: 'radio',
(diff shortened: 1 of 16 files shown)


rvsia (commented):
Code looks good, could we also please add documentation for that functionality? 😊

Issue #1428, opened 2023-12-05
Conditions for fields in field arrays


Hyperkid123:
### Discussed in https://github.com/data-driven-forms/react-forms/discussions/1426

<div type='discussions-op-text'>

<sup>Originally posted by **yldshv** November 30, 2023</sup>
Hello everyone, today I encountered a problem for which there might already be a solution, but I might not be able to find it.

Snippet:
```json
{
   "name":"list",
   "label":"List",
   "component":"object-list", //our name for a field-array
   "fields":[
      {
         "name":"status",
         "label":"Status",
         "component":"select",
         "clearedValue":null,
         "dataType":"integer",
         "options":[
            {
               "label":"Status 1",
               "value":1
            },
            {
               "label":"Status 2",
               "value":2
            },
            {
               "label":"Status 3",
               "value":3
            }
         ],
      },
      {
         "name":"test",
         "label":"Test",
         "component":"radio",
         "dataType":"boolean",
         "options":[
            {
               "label":"No",
               "value":false
            },
            {
               "label":"Yes",
               "value":true
            }
         ],
         "condition":{
            "or":[
               {
                  "when":"list[0].status", //just to highlight this
                  "is":1
               },
               {
                  "when":"list[0].status", //just to highlight this
                  "is":3
               }
            ],
            "then":{
               "visible":true
            }
         }
      }
   ]
}
```

As you can see here, I have list[0].status in the "when" field as a field name. Now, logically, the condition only works for the first item in the array list.

Is there a way to pass this differently in the schema, for example, list[index].status or something similar? If not, wouldn't that be a useful addition?

Thanks in advance and best regards,
Daniel</div>

Hyperkid123:
@yldshv the second commit should fix the condition chaining

Hyperkid123:
New version with a fix was released.

PR #1429, opened 2023-12-05
fix(renderer): add mapped conditions to condition triggers

Hyperkid123:
Ref #1428


Hyperkid123:
Forgot Rich is on vacation.

PR #1440, opened 2024-01-23
fix(all): replace defaultProps with inline function defaults

### LINKED ISSUE #1439: Refactor of defaultProps usages to use default parameters instead
**Scope**: All around the library

**Description**
When using numerous different components across the library, this error is received

Warning: FormTemplate: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.

**Schema**
Any minimal schema which uses a MUI Text box or something will reproduce this

Is there a timeline for this refactoring/will it be occuring?

Hyperkid123:
Fixes #1439 

@rvsia this will need proper validation


packages/blueprint-component-mapper/src/dual-list-select/dual-list-select.js
```diff
@@ -28,7 +28,7 @@ const useStyles = createUseStyles({
   },
 });
 
-const List = ({ value, optionClick, noOptionsTitle, filterValue, filterValueText, selectedValues, MenuProps, MenuItemProps }) => {
+const List = ({ value = {}, optionClick, noOptionsTitle, filterValue, filterValueText, selectedValues, MenuProps = {}, MenuItemProps = {} }) => {
```
rvsia, line 31: value should be array

packages/common/src/form-template/form-template.js
```diff
@@ -96,26 +96,16 @@ FormControls.propTypes = {
   buttonsProps: PropTypes.object,
 };
 
-FormControls.defaultProps = {
-  submitLabel: 'Submit',
-  cancelLabel: 'Cancel',
-  resetLabel: 'Reset',
-  canReset: false,
-  canSubmit: false,
```
rvsia, line 104: is this not used?

Hyperkid123: It's been moved to the `FormControls` argument defaults.

rvsia (approved):

